PR #7, opened 2021-03-03
CONTRIBUTING.md

dylanesque:
Adding CONTRIBUTIING.md file


CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+
+## [Code of Conduct](https://github.com/codeforboston/codeforboston.org/blob/master/code-of-conduct.md)
+
+The Police Misconduct Data Collaborative project has a [Code of Conduct](https://github.com/codeforboston/codeforboston.org/blob/master/code-of-conduct.md)
+to which all contributors must adhere.
+
+## [Issues]()
+
+* [Feature Requests]()
+* [Submitting a Bug Report]()
```
dwreeves, line 16: I think this is supposed to link to something like `https://github.com/codeforboston/police-data-trust/issues/new?assignees=DMalone87&labels=Status%3A+Review+Needed%2C+Type%3A+Bug&template=bug_report.md&title=[BUG]`, which should also work if we use [issue templates](https://docs.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates) which can be a part of the PR.

(I believe feature requests and bug reports are two of Github's default issue templates, so we can just use those default templates.)

dylanesque: Right, I'll fix that tomorrow afternoon.

dylanesque: Ok, this language has been streamlined, and I'm focused on updating the links and adding more specific language changes as the project repo gets more fleshed out.

DMalone87 (approved):

PR #8, opened 2021-03-08
reorganize codebase

dwreeves:
## Refactors to project structure

Some notes and questions regarding what to do in the repo that I think we should work out:

### Project Structure:

- Dumping `.py` files in the root directory isn't going to scale. The API code should be inside a folder called something like `api`, `backend`, or `app`.
- **(Question)** What should this folder be called? I think `backend` is a good name.
- There should be a setup guide for new contributors. Have them install the correct version of Python and install Postgres. It can go into some docs e.g. with mkdocs, or directly into the README. _(note: not done yet)_
- I added continuous integration via Github Actions for both flake8 and pytest.

### Python env stuff:

- **(Question)** Should we be using Pipenv? Or just pip? (See notes at bottom on why I think we should use normal pip)
- **(Question)** Python 3.9 is a bit on the new side. I think we might be a little safer using Python 3.8? WDYT?

### Flask stuff:

- I recommend using a `.flaskenv` file for the `FLASK_APP=?` var.
- I recommend using the "app factory" pattern. https://flask.palletsprojects.com/en/1.1.x/patterns/appfactories/ It has a few benefits involving creating multiple instances of the app, and most experienced Flask devs use this pattern. But moreover it's just organizes the code very nicely in my opinion.
- **(Question)** Can we agree on pytest for unit-testing?

### Style Guide & Style Enforcement:

- We should have a Python style guide in `CONTRIBUTING.md`. It should basically say to defer to the PEP-8 and the Google Python Style Guide. _(note: not done yet)_
- **(Question)** I see a mix of indent lengths and tabs + spaces in the code. Can we agree on the normal PEP8 convention here?
- **(Question)** Max line length-- 80, 100, 120?
- **(Question)** Should we be using flake8? Black?
- **(Question)** Should flake8 be run as part of the continuous integration for PRs?
- We should also have a naming convention for database column names and JSON names. This style preference should be recorded in our style guide.
- **(Question)** How should we do column names? How should we do JSON names?
- My preference is either (1) `snake_case` for both; or (2) `snake_case` for the db + `camelCase` for JSON.

## Thoughts on Pipenv

There are a few problems we are going to face regarding Pipenv. My personal recommendation is use builtin pip and venv, not Pipenv, for the following two reasons:

1. Pipenv requires installing something new on your computer for many users. Code For Boston has high churn, and these little tiny frictions to contributing.

    The mere fact that people need to download something new (outside of a pip install) should be seen as a slight minus for using Pipenv over pip. All else equal, installing less and having less setup is better than installing more and having more setup. (Also adds more bloat to the Github Actions continuous integration for example.)

    To _somewhat_ address the problem, we need to include instructions on how to install Pipenv as part of a "setting up" guide.

2. This next problem is hard to resolve. Basically, OSX and Windows use different dependencies, and Pipenv isn't good at resolving this problem on its own.

    This isn't a theoretical concern. We already have an issue here, i.e. `psycopg2`, which requires installing `psycopg2-binary` on OS X instead. This makes collaborative work extremely annoying. In this case OSX users are on the painful side of the deal, but make no mistake-- Windows users can be on the painful side in the case of packages using C dependencies, such as Meinheld which installs cleanly on OS X and Unix but terribly on Windows.

    The way we resolved this in the Water project was to just use normal pip, and build the frozen environments for both OSX and Windows using pip-tools. https://github.com/codeforboston/flagging The startup scripts for OSX and Windows (`run_unix_dev.sh` and `run_windows_dev.bat`) reference the requisite requirements files using pip. So nobody has an issue. The advantage that pipenv has of separating dev and prod dependencies is also resolved by this system (see our `requirements/` folder to see what I mean.)


dylanesque:
My two cents:

- I fully agree with you on the Project Structure points.
- I think line length should be capped at 80, I don't see what we have to gain from longer lines.
- I think it would reduce mistakes if we stuck to snake casing for both the db and JSON. I did this for a React/Hasura/Apollo project I'm working on to reduce boilerplate for API calls, and I feel like i that choice saved me some unnecessary aggravation.

dwreeves:
An upside of 80 is that it's the "standard" in Python and it's good for open source projects to stick to the standards.

A downside is sometimes you can't fit as much as you'd like on one line. It does come up.

I've also seen 88 as a default for Python linters (e.g. Black uses this as a default). The idea behind 88 is it's 80, but with wiggle room when your code would be prettier going a couple chars over.

Let's gather more feedback on Tuesday. I'm indifferent to the exact number, but I do think we should set one.

dwreeves:
@DMalone87 ready!

DMalone87 (approved):

PR #24, opened 2021-03-31
Folder structure, db commands, other tiny adjustments

dwreeves:
### New folder structure

The major changes are:

* renamed `pull-public-data/` to `scraper/`, which makes importing as a module easier.
* renamed `pull-public-data` to `mapping_police_violence`, which is the name of the site being pulled from in the file.
* Put everything in a `database/` folder. SQLA models go inside `database/models/`.
* Moved the `sql/` folder. Not a good spot for it, since the frontend doesn't need it (so doesn't need to be a shared resource) and in particular the database is what needs it (so it's good to just stick it near the database).

What it looks like:

```
backend/
├── __init__.py
├── api.py
├── celery
│   ├── __init__.py
│   ├── tasks.py
│   └── worker.py
├── config.py
├── database
│   ├── __init__.py
│   ├── core.py
│   ├── models
│   │   ├── __init__.py
│   │   └── incidents.py
│   └── queries
│       └── example_incidents.sql
├── routes
│   ├── __init__.py
│   └── incidents.py
├── scraper
│   ├── __init__.py
│   ├── configs.yaml
│   ├── mapping_police_violence.py
│   └── varnames_crosswalk.csv
├── tests
│   ├── README.md
│   ├── conftest.py
│   ├── test_cli.py
│   └── test_routes.py
└── utils.py
```

### Database CLI commands

I added some commands. You can see them by running `flask psql --help`:

```
Usage: flask psql [OPTIONS] COMMAND [ARGS]...

  Collection of database commands.

Options:
  --help  Show this message and exit.

Commands:
  create        Create the database from nothing.
  delete        Delete the database.
  gen-examples  Generate 2 incident examples in the database.
  init          Initialize the database schemas.
```

These commands have also been added to the run scripts in the root directory.

### Very tiny unit-test adjustments

Added a new test for the shell context processor (hardly that important), and added a bit to the `conftest.py` file regarding database stuff.

### Reorganize `api.py`

Added a nice little structure to `api.py` that I think is much easier to read. The `create_app()` function is now just this:

```python
def create_app(config: Optional[str] = None):
    """Create the API application."""
    app = Flask(__name__)
    config_obj = get_config_from_env(config or app.env)
    app.config.from_object(config_obj)

    with app.app_context():
        register_extensions(app)
        register_commands(app)
        register_routes(app)
        register_misc(app)

    return app
```

### Removed schema initiation from `incidents.sql`

The issue with initializing database schemas in a `.sql` file is that SQLAlchemy already does that for you, and there are good reasons to be doing this process in SQLAlchemy (Alembic, ORM is a good SSoT) rather than manually. So I deleted that.

That said, I think there is a feeling of convenience that is hard to get away from when it comes to writing `.sql` files, and it's really important when we're talking about a volunteer driven project to make contributing as easy as possible. So I added an `execute_query()` command.

There is an example of it in action inside `/backend/database/core.py`:

```python
@db_cli.command("gen-examples")
def gen_examples_command():
    """Generate 2 incident examples in the database."""
    execute_query("example_incidents.sql")
    click.echo("Added 2 example incidents to the database.")
```

### Flask shell context processor

The `flask shell` command is super useful and I've already helped two people out with it. It makes my and others' lives a lot easier in the Flask shell if we already have the `app` and a `client` inside of there to begin with.

### Gunicorn in requirements

Added this to make Heroku deployment possible. It's safe to add to `_core.in` and won't cause problems for either Windows or Unix users.

In the spirit of assisting in Heroku deployment, I also made some adjustments to `config.py`. In particular, the `SQLALCHEMY_DATABASE_URI` is now a property, and for production it's wrapping `os.environ.get("DATABASE_URL")`.


.flaskenv
```diff
@@ -1,2 +1 @@
 FLASK_APP=backend.api:create_app()
-POSTGRES_DB=police_data
```
dwreeves, line 2: My opinion is `.flaskenv` is a pretty bad spot for environment variable names except for `FLASK_APP=?`. The reason why it's not a good spot for stuff is because there should be one source of truth for the environment and the `config.py` is a better spot. I made a mistake by initially putting `POSTGRES_DB` in here, so I'm deleting.

DMalone87 (approved):

PR #29, opened 2021-04-20
Began app.json and added deploy button

dylanesque:
The big question mark to me here is the version of heroku-postgresql we're using, but this should at least be a start


dwreeves:
> The big question mark to me here is the version of heroku-postgresql we're using, but this should at least be a start

Version is optional.

dwreeves:
LGTM

PR #30, opened 2021-04-21
Add Dockerfile, docker-compose config

mikeyavorsky:
Your `.env` file should include (along with any other required vars):
```
POSTGRES_USER=postgres
POSTGRES_PASSWORD=postgres
POSTGRES_DB=police_data
POSTGRES_HOST=db
```

- db initialization script is ./init-user-db.sh, may need to be siloed / organized better
- exposes port 5000 of the api, port 5432 for postgres db by default
- takes POSTGRES_USER POSTGRES_PASSWORD and POSTGRES_DB from .env instead of docker-compose

build and run with `docker-compose build; docker-compose up -d; docker-compose logs -f app`


docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+services:
+  db:
+    image: postgres
+    restart: always
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: postgres
+      POSTGRES_DB: police_data
+    volumes:
+      - ./init-user-db.sh:/docker-entrypoint-initdb.d/init-user-db.sh
```
dwreeves, line 11: I'd use this as the volume `- postgres:/var/lib/postgresql/data`

with this at the end of the file:

```yaml
volumes:
  postgres: {}
```

and I'd also connect this service to the to the dotenv file with:

```yaml
    env_file:
      - ".env"
```

dwreeves:
One note about how the postgres service is defined, but otherwise LGTM

Stampeder525 (approved):
For the first time in world history I can open the site

ryanmahan (dismissed):
Worked for me too

docker-compose.yml
mikeyavorsky: I've added both

dwreeves: sorry it wasn't clear, also replace

```yaml
    volumes:
      - ./init-user-db.sh:/docker-entrypoint-initdb.d/init-user-db.sh
```

with this:

```yaml
    volumes:
      - postgres:/var/lib/postgresql/data
```

and then delete the `init-user-db.sh` file.

This appears to be a more standard way to dockerize Postgres databases that helps to persist data across migrations and deployments.

dwreeves:
One more comment regarding the Postgres service. Once we've resolved that, I'll give a LGTM.

DMalone87 (approved):

mikeyavorsky:
> One more comment regarding the Postgres service. Once we've resolved that, I'll give a LGTM.

@dwreeves According to the Postgres Docker image documentation, the init script and db volume can coexist:
`Warning: scripts in /docker-entrypoint-initdb.d are only run if you start the container with a data directory that is empty; any pre-existing database will be left untouched on container startup.`
See the [initialization scripts](https://hub.docker.com/_/postgres) section.

PR #38, opened 2021-05-25
Second half of db.Models

dwreeves:
I merged in Ryan's PR, made a few minor adjustments, then added in my own models here.


backend/database/models/officer.py
```diff
@@ -0,0 +1,38 @@
...
+class Officer(db.Model):
+    id = db.Column(db.Integer, primary_key=True)  # officer id
+    first_name = db.Column(db.Text)
+    last_name = db.Column(db.Text)
+    race = db.Column(db.Enum(Race))
+    gender = db.Column(db.Enum(Gender))
+    appointed_date = db.Column(db.DateTime)
+    badge = db.Column(db.Text)
+    unit = db.Column(db.Text)  # type?
+    # TODO: How do we handle changes over time in rank?
```
DMalone87, line 29: Each officer object will be related to a single event; I think we should be fine just capturing instantaneous rank in that case. For a combined officer object we can take the most recent rank.

backend/database/models/officer.py
```diff
@@ -0,0 +1,38 @@
...
+    rank = db.Column(db.Text)  # type?
+    star = db.Column(db.Text)  # type?
+    age = db.Column(db.Integer)
+    # TODO: Age changes over time. Might we use birth year?
```
DMalone87, line 33: I agree on using DOB here instead

backend/database/models/incidents.py
```diff
@@ -1,69 +1,144 @@
...
-class Status_Enum(enum.Enum):
-    Unknown = 1
-    Unharmed = 2
-    Injured = 3
-    Disabled = 4
-    Deceased = 5
 
+# TODO: not sure if the below model should be used.
 
 class Incidents(db.Model):
```
ryanmahan, line 143: This is the only model we have that's pluralized. File name is also pluralized
```suggestion
class Incident(db.Model):
```

dwreeves: Yeah that's the old model, and yep the model class names should be singular. I might as well delete it, I guess?

ryanmahan: Oh I see that. Yeah might be a good idea to remove it just to avoid confusion, along with the incidents query in the database/queries folder?

dwreeves:
this PR should be ready to go.

DMalone87 (approved):

Issue #56, opened 2021-06-16
[FEATURE] JEST Unit testing


JMStudiosJoe:
add jest unit testing

PR #57, opened 2021-06-16
started jest unit testing with basic snapshots

JMStudiosJoe:
starting https://github.com/codeforboston/police-data-trust/issues/56


JMStudiosJoe:
Does this look good to frontend devs? Anyone built with [Nextjs](https://nextjs.org/learn/basics/create-nextjs-app) before and have insight?

dylanesque:
I've used Jest with Next before, it's pretty straightforward, and this setup looks correct to me.

PR #72, opened 2021-06-23
understanding routing and adding some pages

JMStudiosJoe:


nancyyang94 (approved):

frontend/tests/snapshots/dashboard.test.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from "react"
+import Dashboard from "../../pages/dashboard"
+import renderer from "react-test-renderer"
+
+it("renders Dashboard correctly", () => {
```
Abhispy007, line 5: Could you explain this part? Looks like "it" is an if statement that is comparing a component tree (what is that exactly) to a snapshot .json file.

JMStudiosJoe: it is the generated HTML output of the react component imported in, `tests/snapshots/ __something.snap.json` i think. Then when the tests are re run any changes made to the component are then compared to the snapshot and most the time you have to keep the snapshots up to date with simple command but make the tests pass, you will see as you work with it.

PR #74, opened 2021-06-29
Starts map component with leaflet (react-leaflet)

sashamaryl:
added leaflet & react-leaflet
added (temp) data sets for stated and county boundaries
added map component which displays state and county boundaries, and selected state for states

added temp map page for developing the UI


docker-compose.yml
```diff
@@ -17,6 +17,7 @@ services:
       dockerfile: ./frontend/Dockerfile
     volumes:
       - ./frontend/pages:/usr/src/app/pages
+      - ./frontend/components:/usr/src/app/components
```
JMStudiosJoe, line 20: does this need to be in this pr?

frontend/components/map.module.css
```diff
@@ -0,0 +1,21 @@
+.mappage {
+  height: 100vh;
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  overflow: hidden;
+}
+
+.mapframe {
```
JMStudiosJoe, line 10: need to make responsive

frontend/components/map.module.css
```diff
@@ -0,0 +1,21 @@
+.mappage {
+  height: 100vh;
+  width: 100%;
+  display: flex;
+  flex-direction: column;
```
JMStudiosJoe, line 5: lets use css-grid instead of flex much easier to understand

frontend/pages/map.tsx
```diff
@@ -0,0 +1,5 @@
+import dynamic from "next/dynamic"
+
+// Temporary page to display the leaflet map for development
+
+export default dynamic(() => import("../components/Map"), { ssr: false })
```
JMStudiosJoe, line 5: lets move this to the dashboard page since that is where the map will live :)

JMStudiosJoe (changes requested):
fantastic work thank you very much, just some minor comments! We should discuss with @DMalone87  and see how React Leaflet feels. I was able to discuss this app with our Open Street Maps team and they were saying for some map like this it would be best to use D3 for the starting zooming into location whereas Leaflet might be a little too clunky for that part. Ultimately we should run the feature by @DMalone87 and see what he thinks.

frontend/pages/map.tsx
DMalone87: Including the link to D3 here for the sake of discussion. I'll test leaflet out and give some feedback in the next few days.
https://observablehq.com/@d3/gallery

dylanesque: Dropping a link to a choropleth I made using d3 so folks can see how it works: https://github.com/dylanesque/Choropleth

docker-compose.yml
sashamaryl: this line is for code reloading in Docker, and goes with the component folder, which I'm adding in this pr -- so I think yes

frontend/components/map.module.css
sashamaryl: (I'm on board -- just going to drag my toes a bit on the layout items until we decide which tool we'll use and are building the real deal map)

frontend/pages/map.tsx
sashamaryl: Thanks all for your comments! See you tonight!

frontend/components/map.module.css
srsexton94: We talked about this in my PR & it seemed the consensus was to use css-grid for multi-dimensional layouts and flexbox for single-dimensional/content-based layouts 
(aligns with the intended differences between them - here's an article I found helpful! https://css-tricks.com/css-grid-replace-flexbox/)

frontend/components/Map.jsx
```diff
@@ -0,0 +1,117 @@
...
+            </div>
+        </div>
+    )
+}
+
+export function GeoDataStates() {
+
+    const map = useMap()
+
+    const [gState, setGState] = useState()
```
srsexton94, line 34: I've always heard the advice to avoid abbreviations in variable names as it helps with clarity (like as a first-time reader I don't know what 'G' stands for)

frontend/components/Map.jsx
```diff
@@ -0,0 +1,117 @@
...
+import styles from './map.module.css'
+
+export default function Map() {
+
+    let center = [35, -90]
+    let startingZoom = 4
+    const [selectedState, setSelectedState] = useState(1)
+
+    return (
+        <div className={styles.mappage} >
```
srsexton94, line 15: If you like, I've found it helpful to deconstruct the applicable `className`s from `styles` (so that you don't have to write `styles.` repeatedly!)

frontend/components/Map.jsx
```diff
@@ -0,0 +1,117 @@
+import 'leaflet/dist/leaflet.css'
```
srsexton94, line 1: I think we've been using `tsx` over `jsx` if you're comfortable!

srsexton94: Also, wondering if we might want to break these out into their own component files?

frontend/components/Map.jsx
```diff
@@ -0,0 +1,117 @@
...
+    )
+}
+
+export function GeoDataStates() {
+
+    const map = useMap()
+
+    const [gState, setGState] = useState()
+
+    const onEachFeature = (feature, layer) => {
```
etilden, line 36: I think this function could be more descriptively names-- to me this reads as "this function will do ___ when you click on a state". I think it would be helpful for the name to describe what the function will do.

etilden: (same for onEachFeatureCounty)

JMStudiosJoe:
lets fix the merge conflicts then can get this pr in next :)

sashamaryl:
Hey all! Thanks so much for your comments!! 
This PR is the old one with leaflet map that I mentioned the other day, and shouldn't be merged. We decided a couple weeks ago that we were going to go with D3 instead. I just made the new PR starting the map using d3 #99  and we should move forward with that one. I will close this one now.

PR #75, opened 2021-06-29
Start header component, basic structure and positioning.

nancyyang94:
- in progress: fixing banner width, need to make responsive.


dylanesque (commented):
This looks very good overall, but I think there's some room for improvement as far as semantic HTML goes. The wrapping div for the component would be more accessible if it was a `<header>`

nancyyang94:
> This looks very good overall, but I think there's some room for improvement as far as semantic HTML goes. The wrapping div for the component would be more accessible if it was a `<header>`

Thank you for the feedback - I will update the wrapping div to be a <header>

nancyyang94:
7/6 updates:
1) semantic HTML changes - updated wrapper div to be `<header>,` updated links to be `<nav>`
2) design updates - implemented white tab display for currently selected nav item (have not added links yet), margin changes, donate button added

WIP:
3) css grid layout - changed displays from flex to grid per feedback last week. i used px for the column widths from figma, would it be better to calculate these as percentages of the total px width instead? below is the current composition:

image

4) snapshot tests

JMStudiosJoe:
still a WIP? for 3) percentages can get tricky but rem and em do a little bit of that for you as well. Looking great so far

frontend/compositions/header/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+export default function Header() {
+  let [selected, setSelected] = useState('Search');
+
+  const handleNavChange = (e: any) => {
+    let value = e.target.innerText;
+    setSelected(value);
+  }
+
+  return (
+    <header className={styles.wrapper}>
```
srsexton94, line 15: If you like, something I've found helpful is to deconstruct the `classNames` from `styles` so that you don't have to write `styles.` in each of them!

Total non-blocker tho!

frontend/compositions/header/header.module.css
```diff
@@ -0,0 +1,85 @@
+.wrapper {
+  width: 100%;
+  height: 124px;
```
srsexton94, line 3: Can we use `rem` over `px`? Important for zoom responsiveness for low vision users

If it merges ahead of you, feature58 will introduce CSS size variables that will manage that conversion. 
But if not, or if you'd like to just do it now, basically divide your `px` number by 16 & that'll be your `rem` number!

frontend/compositions/header/index.tsx
nancyyang94: @srsexton94 this is super helpful - thank you for suggestion!

frontend/compositions/header/header.module.css
nancyyang94: @srsexton94 yes I will convert from px to rem! thank you for your review + feedback.

frontend/compositions/header/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+        </div>
+        <div className={styles.titleContainer}>
+          <h2 className={styles.title}>National Police Data Coalition</h2>
+          <p className={styles.subtitle}>The national index of police incidents</p>
+        </div>
+        <nav>
+          <ul className={styles.nav}>
+            <li className={selected === 'Search' ? styles.tab : ''} onClick={handleNavChange}>Search</li>
+            <li className={selected === 'Profile' ? styles.tab : ''} onClick={handleNavChange}>Profile</li>
+            <li className={selected === 'About' ? styles.tab : ''} onClick={handleNavChange}>About</li>
```
srsexton94, line 28: We could factor out the classNames into a helper function to DRY this up a bit!

frontend/compositions/header/index.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { useState } from "react"
```
srsexton94, line 1: Could we rename this component to be more specific (maybe `DashboardHeader`?) To differentiate it from other "headers" we have throughout the app

frontend/compositions/header/index.tsx
srsexton94: Also! We'll need to make these a natively focusable element for it to be accessible to keyboard users

We'll probably want to review our routing strategy as a team, in the meantime adding anchor tags (inside of each `li` and around their text) as a placeholder is probably sufficient!

(in that case, we'll also want to move `handleNavChange` to the link click instead of the list item click)

frontend/compositions/header/header.module.css
```diff
@@ -0,0 +1,85 @@
...
+  height: 25px;
+  border-radius: 3px 3px 0px 0px;
+  box-shadow: 0px -2px 4px rgba(48, 52, 99, 0.25);
+}
+
+.donate {
+  margin-top: 21px;
+}
+
+.donate button {
```
srsexton94, line 73: I'm curious why we wrapped this in the `.donate` div instead of applying styles directly to the button!

frontend/compositions/header/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+import NPDCLogo from './assets/NPDCLogo.svg'
+import styles from './header.module.css'
+
+
+export default function Header() {
+  let [selected, setSelected] = useState('Search');
+
+  const handleNavChange = (e: any) => {
```
srsexton94, line 9: I believe we want to avoid `any` as a type - I believe the type you'll need here is `MouseEvent<HTMLAnchorElement>` (once we add the links at least - and `MouseEvent` comes from react)

srsexton94 (changes requested):
Here are a few things I was thinking over!

(I also couldn't seem to leave a file-level comment but I imagine we'll want to import images from the `public/` directory over `compositions/`)

srsexton94:
Hey! I can't seem to find your branch to pull - but the screenshot makes it seem like the page tabs and donate button are about halfway across the page - design-wise we intended them to be on the right side of the page (apologies if that wasn't clear in the wireframes!

frontend/compositions/header/index.tsx
nancyyang94: @srsexton94 thank you for all the feedback! I've added anchor tags with a "#" placeholder link for the meantime inside. of each `li` tag. 

I moved the full nav bar and the handleNavChange function into a separate component, and will factor out classNames into a helper function as well.

frontend/compositions/header/index.tsx
nancyyang94: Yes, will rename!

frontend/compositions/header/header.module.css
nancyyang94: Thank you for the catch  -  will apply styles to button directly.

frontend/compositions/header/index.tsx
nancyyang94: > Hey! I can't seem to find your branch to pull - but the screenshot makes it seem like the page tabs and donate button are about halfway across the page - design-wise we intended them to be on the right side of the page (apologies if that wasn't clear in the wireframes!

Hi Sam, thank you for all the helpful feedback! 

I'll push up my latest changes tonight, with a change to the grid layout so that the page tabs and donate button stay on the right side. The wireframes are totally clear - I was incorrectly setting up my columns, and will make sure to attach screenshots of the fix!

nancyyang94:
**7/12 Updates:**

1) Created mobile view - refactored CSS to use mobile design as default, added media query for laptop breakpoint (I set this at 1100px/68.75rem per the global stylesheet, but not sure if we agreed upon set breakpoints?). Converted all px units to rem.

2) Nav bar updates - factored out desktop nav bar to be separate component, added anchor tags, DRYed up code for setting class names.

3) Media query hook - I added a media query hook in helperFunctions, to conditionally render different logo/text/nav bar displays based on screen-size, in index.tsx. Any feedback on best practice for this would be much appreciated - would it be better to instead use CSS to display/hide for this smaller composition? 

4) Renamed composition to more specific DashboardHeader.

5) Fixed grid template columns so that the left logo + title stay together on the left, and the right nav bar + donate button stay together on the right.

**Screenshots:**

Mobile View (375)

DashboardHeaderMobileView

Desktop View (1440) 
DashboardHeaderDesktopView

frontend/compositions/dashboardHeader/desktopNav.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const handleNavChange = (e: React.MouseEvent<Element>) => {
+    const value = (e.target as HTMLElement).innerText;
+    setSelected(value);
+  }
+
+  const setClassName = (linkName: string) => {
+    if (linkName === selected) {
+      return styles.tab;
+    }
+    return '';
```
srsexton94, line 16: We could use a ternary to avoid duplicate return statements it you like! 
ala `return linkName === selected ? styles.tab : ''`

frontend/compositions/dashboardHeader/header.module.css
```diff
@@ -0,0 +1,156 @@
+/* default - mobile styles */
+.wrapper {
+  width: 100%;
+  height: 6.94rem;
```
srsexton94, line 4: If you don't mind implementing - we have new global sizing variables in `styles/variables.css`! 
It maps the `px` values to their appropriate `rem` (ie here you could use `var(--size112)` to map to `7rem`)

frontend/compositions/dashboardHeader/desktopNav.tsx
srsexton94, line 16: We could use a ternary to avoid duplicate return statements if you like!
ala `return linkName === selected ? styles.tab : ''`

frontend/compositions/dashboardHeader/header.module.css
```diff
@@ -0,0 +1,156 @@
...
+  display: flex;
+  flex-direction: column;
+  align-items: flex-end;
+}
+
+.nav {
+  margin: 1.69rem 1.69rem 1.8rem;
+}
+
+.button {
```
srsexton94, line 63: There's a new global class for `primaryButton` you could use here if you like! You might have to specify sizing still but might DRY up the rest!

frontend/compositions/dashboardHeader/helperFunctions.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export function useMediaQuery(width: number): boolean {
+  const [targetReached, setTargetReached] = useState(false);
+
+  const updateTarget = useCallback((e) => {
+    if (e.matches) {
+      setTargetReached(true);
+    } else {
+      setTargetReached(false);
+    }
```
srsexton94, line 11: We could simplify this as `useCallback((e) => { setTargetReached(e.matches) }, [])` if you like!
(or `!!e.matches` if "matches" isn't a boolean)

frontend/compositions/dashboardHeader/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+          <p>The national index of police incidents</p>
+        </div>
+      </div>
+
+      <div className={rightHeader}>
+        <nav className={nav}>
+          {navBar}
+        </nav>
+        <button className={button} type="button">
+          <p>DONATE</p>
```
srsexton94, line 43: If we nest the button text in a paragraph element it won't be picked up appropriately as the accessible name of the button for screen reader users - I think we should be able to just write the text in directly!

frontend/compositions/dashboardHeader/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const belowBreakpoint = useMediaQuery(375);
+  const logo = belowBreakpoint ? <LogoMobile /> : <NPDCLogo />;
+  const title = belowBreakpoint ? 'N.P.D.C.' : 'National Police Data Coalition';
+  const navBar = belowBreakpoint ? <HamburgerMenu /> : <DesktopNav />
+
+  return (
+    <header className={wrapper}>
+      <div className={leftHeader}>
+        <div className={logoContainer}>
```
srsexton94, line 29: We have a `Logo` component now you can use! at `shared-components/logo.tsx` (you'll just have to affect the sizing)

frontend/compositions/dashboardHeader/helperFunctions.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { useState, useCallback, useEffect } from "react"
```
srsexton94, line 1: We have a `helpers/` directory now (at the frontend root level) if you wanted to move this file there!

frontend/pages/dashboard/index.tsx
```diff
@@ -1,3 +1,15 @@
+import DashboardHeader from '../../compositions/dashboardHeader/index';
+import styles  from '../../compositions/dashboardHeader/header.module.css';
```
srsexton94, line 2: I've always been warned not to apply other components' local styles to other components - seems like `backgroundBanner` might best live inside the `DashboardHeader` component directly?

srsexton94: Also I think we'll want to rename the css file to match the new component name!

srsexton94 (changes requested):
Looking good! Great job on the mobile view! 
Some comments re: consistency with recent merge & code conventions - but impressive all around!

Also, wasn't sure where to comment but I think we'll want to type inputs/outputs of functions wherever it's not implicitly inferred (ie if you hover over it, does it display the correct type?)  
That's not super common though so most inputs/outputs should have a type (if it's not returning anything, the output type is `void`)

frontend/compositions/dashboardHeader/header.module.css
nancyyang94: Apologies for the late response - will convert all rem to global sizing variables!

frontend/compositions/dashboardHeader/helperFunctions.tsx
dylanesque: A small change I would suggest for maintainability is changing the import style here so it's not default importing React proper, it should be something more like `import * as React, { useState, useCallback, useEffect } from "react"`. Thos blog post explains the rationale behind that: https://epicreact.dev/importing-react-through-the-ages/

frontend/pages/dashboard/index.tsx
nancyyang94: Thanks for the catch - I've removed background banner from the dashboard component all together, and lives inside DashboardHeader directly. CSS file renamed as well!

dylanesque:
> 3. Media query hook - I added a media query hook in helperFunctions, to conditionally render different logo/text/nav bar displays based on screen-size, in index.tsx. Any feedback on best practice for this would be much appreciated - would it be better to instead use CSS to display/hide for this smaller composition?

The media query hook in this case seems fairly heavyweight to me. Memoizing hooks like `useMemo` or `useCallback` should ideally only be used to fix problems of referential integrity or computationally expensive operations. CSS should more than do the job for resizing here.

frontend/compositions/dashboardHeader/helperFunctions.tsx
nancyyang94: Good to know - thank you for sharing the blog post as well for more context.

nancyyang94:
> > 1. Media query hook - I added a media query hook in helperFunctions, to conditionally render different logo/text/nav bar displays based on screen-size, in index.tsx. Any feedback on best practice for this would be much appreciated - would it be better to instead use CSS to display/hide for this smaller composition?
> 
> The media query hook in this case seems fairly heavyweight to me. Memoizing hooks like `useMemo` or `useCallback` should ideally only be used to fix problems of referential integrity or computationally expensive operations. CSS should more than do the job for resizing here.

Thank you for the feedback - I was concerned that I was overcomplicating this, and that makes a lot of sense to use those hooks sparingly. I will plan to refactor using CSS only.

If it may be at all useful for other components in the future, I can move the media query hook to the root level `helpers/` directory per @srsexton94's comment above? Open to any thoughts, or I can go ahead delete this helper function entirely.

nancyyang94:
> Looking good! Great job on the mobile view!
> Some comments re: consistency with recent merge & code conventions - but impressive all around!
> 
> Also, wasn't sure where to comment but I think we'll want to type inputs/outputs of functions wherever it's not implicitly inferred (ie if you hover over it, does it display the correct type?)
> That's not super common though so most inputs/outputs should have a type (if it's not returning anything, the output type is `void`)

Thank you Sam for all the feedback and helpful pointers. 

I want to make sure I'm understanding/on the right track for typing inputs/outputs of functions where it's not implicitly referred - for example, I added output types to functions in `desktopNav.tsx` as follows: 

` const handleNavChange = (e: React.MouseEvent<Element>) : void => {
    const value = (e.target as HTMLElement).innerText;
    setSelected(value);
  }

  const setClassName = (linkName: string) : string => {
    return linkName === selected ? styles.tab : '';
  }
`
For the dashboardHeader component itself, I updated the output type to `JSX.Element`, but not sure if that may be unnecessary/implicitly inferred.

nancyyang94:
> > Looking good! Great job on the mobile view!
> > Some comments re: consistency with recent merge & code conventions - but impressive all around!
> > Also, wasn't sure where to comment but I think we'll want to type inputs/outputs of functions wherever it's not implicitly inferred (ie if you hover over it, does it display the correct type?)
> > That's not super common though so most inputs/outputs should have a type (if it's not returning anything, the output type is `void`)
> 
> Thank you Sam for all the feedback and helpful pointers.
> 
> I want to make sure I'm understanding/on the right track for typing inputs/outputs of functions where it's not implicitly referred - for example, I added output types to functions in `desktopNav.tsx` as follows:
> 
> ` const handleNavChange = (e: React.MouseEvent) : void => {
> const value = (e.target as HTMLElement).innerText;
> setSelected(value);
> }
> 
> const setClassName = (linkName: string) : string => {
> return linkName === selected ? styles.tab : '';
> }
> `
> 
> For the dashboardHeader component itself, I updated the output type to `JSX.Element`, but not sure if that may be unnecessary/implicitly inferred.

Please disregard earlier comment on return types - feeling good about this moving forward with the clarification in today's meetup!

JMStudiosJoe (approved):

frontend/compositions/dashboardHeader/header.module.css
nancyyang94: Updated to use the global class

frontend/compositions/header/index.tsx
nancyyang94: hi @srsexton94 - i want to double check my understanding of making each `li`  natively focusable - my current understanding is that adding the anchor tags addressed this, i'm seeing that i'm able to tab through the links in sequential order. would it be best to also add a tabindex of 0 to all `li` elements, outside of the `a` tags?

PR #78, opened 2021-07-06
Create Passport Registration

dylanesque:
WIP, this was getting a little lengthy, I wanted to check in with current progress before adding invalid form state styling, DRYing up the CSS, and adding a snapshot test.

Edit: This latest version includes a fix for testing setup (it wasn't working previously) and a refactor to utilize Grid for the form layout.

Edit 2: Final revision? The one thing I'm unhappy with here is the button feels uncannily off from the center a bit. Widening the width of the grid container helped a little, but it looks "wrong" to me. I'm not sure how to fix this apart from removing the button and the link directly underneath it from out of the grid.


frontend/components/StateSelect/UsStateSelect.tsx
```diff
@@ -0,0 +1,59 @@
+import styles from './select.module.css'
+
+export function USStateSelect() {
+    return (
+      <select id="state" className={styles.input}>
```
JMStudiosJoe, line 5: out of curiosity why not use a library for this?

frontend/pages/register/index.tsx
```diff
@@ -1,3 +1,12 @@
+import Link from "next/link"
+
 export default function Register() {
-  return <div className="Register">I AM REGISTRATION!!!!!!</div>
+  return (
+    <>
+      <div className="Register">I AM REGISTRATION!!!!!!</div>
+      <Link href="/register/passport">
```
JMStudiosJoe, line 7: I am pretty sure the flow is always the user must register as a basic user then from there once they are logged in they can then request to update to a passport user or behyond.

frontend/pages/register/passport/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <p>
+          Hello <em>{name}</em>, thank you for your continued interest in the National Police Data
+          Coalition.
+        </p>
+        <p>
+          We are able to provide access to legally protected data to users with the appropriate
+          permissions. This form will submit your profile for approval.
+        </p>
+      </div>
+      <form>
```
JMStudiosJoe, line 25: no need for form

frontend/pages/register/passport/passport.module.css
```diff
@@ -0,0 +1,62 @@
...
+  align-self: center;
+  background-color: var(--darkBlue);
+  color: white;
+  font-size: 19px;
+  height: var(--spacing8);
+  margin-bottom: 0.5em;
+  width: var(--spacing10);
+}
+
+.container {
```
JMStudiosJoe, line 11: I am not sure about the .button class naming but could we switch from css flex to css grid? its far more declarative and descriptive for what it is actually doing https://css-tricks.com/snippets/css/complete-guide-grid/

JMStudiosJoe (changes requested):

frontend/pages/register/index.tsx
JMStudiosJoe: This comment may be disregarded since i see the PR for basic user registration so maybe just need to link this one from the dashboard?

frontend/pages/register/passport/passport.module.css
dylanesque: Yup, I can refactor that out, especially since I've seen the mobile mocks and have a clearer idea of what this should look like on all screens.

frontend/components/StateSelect/UsStateSelect.tsx
dylanesque: For the forms, or the state selection component specifically? At a glance, I'm unsure if pulling in a library for one component justifies the extra dependency we can avoid by just having this pasted in.

frontend/pages/register/index.tsx
dylanesque: It's just a placeholder for me to test that routing was working as expected, it certainly doesn't need to stay here.

frontend/pages/register/passport/index.tsx
dylanesque: So, I don't agree with this. Even if it passes a11y checks on Lighthouse, using some other container is much worse from a semantic point of view. This component takes several pieces of input from the user and will (eventually) send it to the back-end, there is no better role for it than that of form.

JMStudiosJoe: That is all true however from my experience form element has introduced too many side effects and doesn't need to follow the semantic point of view for the risk of side effects involved. We have the component already maintaining the data entered in and do not need the HTML inputs and forms to rely on storing the data, so it fits another separation of concerns by keeping all the core logic in the JS and HTML only for structure.

I am willing to discuss further and we can bring this to the rest of the team as well and see their thoughts, but in my experience using the form element, even for the semantic benefit, just adds too much complexity for the simplicity of the component. Please let me know if you want to further discuss since I will need some convincing on this subject and always a great opportunity to div into something deeper for us to learn. :)

dylanesque: To be clear, is your objection to having a <form> tag in the codebase period, or are you assuming that I'm writing this with the attention of submitting the input data via that form element? Because if it's the latter, that's not my intention.

JMStudiosJoe: hmmm So then why use the form? I think my objective aligns with not using <form> at all

dylanesque: For semantic and a11y reasons, which going against creates more problems for us than it would solve.

JMStudiosJoe: oh to use <form> for accessibility reasons (a11y?) hmmm after doing some googling (https://gomakethings.com/why-use-a-form-element-when-submitting-fields-with-javascript/) looks like you are on the right track and they are more important than I thought. Just to clarify, will only be using for semantic purposes and for a11y reasons and not going to use the built in other stuff with it? In the past that is where I have run into too many issues with using a form but if just for semantic use and a11y bare minimum stuff then sounds good to me.

JMStudiosJoe: Great discussion!

dylanesque: Yeah, I'm not going to say I'll NEVER submit code that isn't idiomatic to the React way of doing things, but that's pretty high on my list of what I think about when writing. Submitting data via form elements is definitely on the list of "don't do this".

JMStudiosJoe (dismissed):

frontend/components/StateSelect/UsStateSelect.tsx
srsexton94: Just un-resolving in case you missed my comment!

frontend/components/Logo/logo.module.css
```diff
@@ -0,0 +1,4 @@
+.logo {
+  margin: 0 auto;
+  width: 160px;
```
srsexton94, line 3: I think we're using `rem` over `px` - but my last merge introduced some variables you can use - `var(--size160)` will map to the right `rem` value for 160px!

frontend/pages/register/passport/index.tsx
```diff
@@ -0,0 +1,95 @@
+import Link from "next/link"
+
+import { USStateSelect } from "../../../components/StateSelect/UsStateSelect"
+import { Logo } from "../../../components/Logo/Logo"
+import styles from "./passport.module.css"
+
+  function handleSubmit() {
```
srsexton94, line 7: Curious why this is outside of `Passport`!

srsexton94 (changes requested):
I think this ticket has scope-crept pretty significantly - my understanding was this was a frontend ticket to create the UI of the Passport application page - but it seems only 4-7 of the 50 file changes have to do with that task

I think it'd probably be a good idea to break out the extra work you did into their own tickets/branches so that it's clear what we're trying to accomplish with each & so that the appropriate people can review your PR successfully

Screenshots/explanatory comments could also be helpful if it's not too much trouble!

frontend/components/Logo/logo.module.css
dylanesque: Yeah, I was actually waiting for your branch to get merged so I could fold in any styling you added that overlapped, and also see what conventions you were using to hew closer to those.

frontend/components/StateSelect/UsStateSelect.tsx
dylanesque: Maybe? 

> I think this ticket has scope-crept pretty significantly - my understanding was this was a frontend ticket to create the UI of the Passport application page - but it seems only 4-7 of the 50 file changes have to do with that task
> 
> I think it'd probably be a good idea to break out the extra work you did into their own tickets/branches so that it's clear what we're trying to accomplish with each & so that the appropriate people can review your PR successfully
> 
> Screenshots/explanatory comments could also be helpful if it's not too much trouble!

Something is very wrong, there are a bunch of changes I didn't make, and I'm not sure why they're there. This will hopefully be more normalized once I pull in the latest changes from master.

frontend/pages/register/index.tsx
```diff
@@ -1,3 +1,12 @@
...
+        <a>Register for a Passport account</a>
+      </Link>
```
srsexton94: Is this also outdated perhaps? I see in the thread above that this was meant to be a placeholder && it should've been replaced with `pages/viewer-registration/viewer-registration.tsx` with the last merge

frontend/components/StateSelect/select.module.css
```diff
@@ -0,0 +1,3 @@
+.input {
```
srsexton94, line 1: Same comment as the other css file re: context-specific & user-facing naming!

frontend/components/Logo/Logo.tsx
```diff
@@ -0,0 +1,5 @@
+import Image from "next/image"
```
srsexton94, line 1: We already have a Logo component at `shared-components/logo.tsx` (& it is/should be sized 144x113 as per wireframes)

srsexton94 (changes requested):
Re-reviewed after changes - seems like there may have been some confusion (ie duplication/reverting) resolving merge conflicts? Happy to help resolve those if you like!

frontend/styles/globals.css
```diff
@@ -178,7 +177,12 @@ body {
   font-size: 18px;
```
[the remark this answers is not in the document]
srsexton94: It does appear to be so! Apologies that "Outdated" flag wasn't on it when I was reviewing

frontend/pages/register/passport/passport.module.css
```diff
@@ -0,0 +1,62 @@
...
+.input {
+  margin-right: var(--spacing3);
+  max-width: 200px;
+}
+
+.label {
+  color: var(--darkBlue);
+  font-weight: 500;
+  margin-right: var(--spacing2);
+  margin-bottom: 0.5em;
```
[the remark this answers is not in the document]
srsexton94: Here's an article about it I found helpful!
https://engageinteractive.co.uk/blog/em-vs-rem-vs-px

frontend/pages/register/index.tsx
srsexton94: Also appears to be so! Apologies, the "file changes" tab doesn't feel super intuitive for review

dylanesque:
Regarding CSS classes: One of the nicer perks of CSS modules (that I wasn't aware of until this afternoon) is that the first part of the filename gets namespaced onto the class name in the browser, so these names are less sparse than they appear on paper.

dylanesque:
@srsexton94 sr So, for the case of the state select component, I think making it a list to map through and render is more WET than anything else. Consider that:

- The list of states will never change from the back-end.
- The list will also never change from the front-end.

Given that, I don't see how adding JS to do a job that HTML is more than up to the task of helps us in this case.

frontend/shared-components/state-select/UsStateSelect.tsx
```diff
@@ -0,0 +1,59 @@
+import styles from "./state.module.css"
+
+export function USStateSelect() {
+    return (
+      <select className={styles.select}>
+        <option value="AL">AL</option>
```
srsexton94, line 6: I still think we should iterate through a string array of states to generate these more DRY-ly!

Something like `{ states.map(state => <option key={state} valued={state}>{state}</option>) }` in place of all the options (with `states: string[] = ['AL', 'AK', 'AR', etc....]`)

EDIT: I only just saw your response, but couldn't add a reply outside of a github "review" - but the above is much DRYer because you're not writing the option element 50 times over (literally 😅) 
It doesn't matter that the array won't change, it can be a constant - the point is just that it's just so repetitive

frontend/pages/passport-application/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </div>
+      <form className={styles.form}>
+        <div className={[styles.address, styles.inputwrapper].join(" ")}>
+          <label htmlFor="address" className={styles.label}>
+            <strong>Street Address:</strong>
+          </label>
+          <input
+            type="text"
+            id="address"
+            placeholder="Enter your street address"
```
srsexton94, line 36: Hey! So thanks for removing the Bob Boberton stuff 😅 but the wireframes actually don't have any placeholder text at all & that's intentional as it's considered inaccessible/bad UX

Here's an article I found helpful on why that is! 
https://www.deque.com/blog/accessible-forms-the-problem-with-placeholders/

frontend/pages/passport-application/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { USStateSelect } from "../../shared-components/state-select/UsStateSelect"
+import styles from "./passport.module.css"
+
+function handleSubmit() {
+  // TODO: add form submission logic
+}
+
+export default function Passport({ name = "Herbert Q. Placeholder" }) {
+  return (
+    <div className={styles.wrapper}>
```
srsexton94, line 13: If you like, on other components we've been destructuring the classNames from the `styles` object (especially when there's more than like 1-3 classNames) 

It helps DRY up the template code a bit not having to write `styles.` repeatedly!

frontend/pages/register/index.tsx
```diff
@@ -0,0 +1,7 @@
+export default function Register() {
```
srsexton94, line 1: Was this potentially added back in re: merge conflicts? This file shouldn't exist anymore - it was replaced with `viewer-registration`

srsexton94 (changes requested):
Just a final few comments! Left a response re: the states dropdown in the file

Also re: the classNames thing - even though CSS modules are name-spaced to the component my suggestion to use more informative names was more just for developer clarity - definitely not a blocker but I think there's a happy middle group to reached between specificity & brevity!

JMStudiosJoe:
can we fix the conflicts @srsexton94 and then would it be good to be reviewed and merged?

dylanesque:
Regarding the state component: Refactoring would be a lot more work since all I did was pull that HTML snippet off the web and plop it in the codebase: it doesn't get DRYer than "CMD + C, CMD + V, using working code for a common use case in forms". Changing that would be more work which makes the code there marginally less efficient since JavaScript is a render-blocking resource. Since the list of states never, ever changes, using that snippet makes sense because we're not trying to reinvent the wheel.

srsexton94:
Hmm, I'm wondering if this state disagreement is coming from a misunderstanding of what DRY code means. [Here's a good/quick overview from one of my favorite clean-code gurus, Uncle Bob :)](https://youtu.be/7EmboKQH8lM?t=5828)

DRY doesn't have anything to do with  HOW you got the code (ie copy-pasted or original). It's about the code itself. The state solution you have is undeniably pretty repetitive code as It repeats the same code structure 50 times. 
I also feel strongly that falling back on "don't reinvent the wheel" (or its nefarious cousin "if it ain't broke don't fix it") can more often than not just act as an excuse to leave messy code (which I 100% know is not your intent!)

I of course didn't mean to make you feel like you had to do extra work though - to help relieve some of that workload here it is written in full - you should be able to just copy/paste this in!
```ts
const states: string[] = [
  "AL", "AK", "AR", "AZ", "CA", "CO", "CT", "DC", "DE", "FL", "GA", "HI", "IA", "ID", "IL", "IN", "KS", 
  "KY", "LA", "MA", "MD", "ME", "MI", "MN", "MO", "MS", "MT", "NC", "NE", "NH", "NJ", "NM", "NV", "NY", 
  "ND", "OH", "OK", "OR", "PA", "RI", "SC", "SD", "TN", "TX", "UT", "VT", "VA", "WA", "WI", "WV", "WY"
]

export function USStateSelect() {
  return (
    <select className={styles.select}>
      { states.map(state => <option key={state} valued={state}>{state}</option>) }
    </select>
  )
}
```

frontend/README.md
```diff
@@ -15,7 +15,7 @@ if you run into an error such as
 Error: .git can't be found
 ```
 
-you can copy the .git/ directory in the root of the project and move it to the frontend/ and try the install again and it should succeed. try the install again and remember do not commit this .git/ you may want to delete it afterwards until figured out alernative.
+you can copy the .git/ directory in the root of the project and move it to the frontend/ and try the install again and it should succeed. try the install again and remember do not commit this .git/ you may want to delete it afterwards until figured out alternative.
```
srsexton94, line 18: Typo correction never fails to make my heart happy 👏🏻🥳🙌🏻

frontend/pages/passport-application/index.tsx
```diff
@@ -1,3 +1,106 @@
...
+          <input type="number" id="zip" required={true} className={zipInput} />
+        </div>
+        <div className={[ppTextarea, inputWrapper].join(" ")}>
+          <label htmlFor="textarea" className={ppLabel}>
+            <strong>Why are you signing up to the NPDC?:</strong>
+          </label>
+          <textarea
+            id="textarea"
+            className={ppTextarea}
+            defaultValue="I want to sign up for the NPDC because..."
```
srsexton94, line 88: Apologies if not clear, this was part of the placeholder-text removal comment from before!

frontend/pages/passport-application/index.tsx
```diff
@@ -1,3 +1,106 @@
...
+      </div>
+      <form className={ppForm}>
+        <div className={[ppAddress, inputWrapper].join(" ")}>
+          <label htmlFor="address" className={styles.label}>
+            <strong>Street Address:</strong>
+          </label>
+          <input
+            type="text"
+            id="address"
+            required={true}
```
srsexton94, line 56: This differs fairly significantly from the error-validation patterns used in Login & ViewerRegistration - BUT imho the error validation stuff felt like it scope-crept wayyy too far out of a reasonable size for these mainly-UI-based tickets

Definitely not a blocker but I think we might want to create a follow-up ticket to work on error validation between the tree "enrollment" type pages!

frontend/pages/passport-application/index.tsx
```diff
@@ -1,3 +1,106 @@
...
+    ppRedirect,
+    ppTextarea,
+    ppButton,
+    ppAddress,
+    ppAddInput,
+    ppZip,
+    ppState,
+    ppCity,
+    zipInput,
+    ppLabel
```
srsexton94, line 30: Not a blocker, but if you wanted to reduce the length here you could utilize CSS nesting to not need quite so many classNames from the module!

srsexton94 (dismissed):

srsexton94:
Hey folks!  Mike & I talked offline - he's going to post some last changes & I've otherwise approved - please leave it for him to merge himself though to make sure everything's ready :)

dylanesque (commented):
Merging and closing PR

> Hey folks! Mike & I talked offline - he's going to post some last changes & I've otherwise approved - please leave it for him to merge himself though to make sure everything's ready :)

Final changes are in, including a satisfactory fix for the button centering, this just needs approval from someone who isn't me :)

srsexton94 (approved):

PR #82, opened 2021-07-07
feature58 ✨ Basic User Registration page

srsexton94:
## This ticket will...
Components
* Create a `ViewerRegistration` component and sets it to `/register` route
* Update `UserLogin` component to use new shared components and set it to `/login` route
* Add error message styling & basic form validation to both (but not submission)

Styles
* Factor out the global stylesheet to gather alike-purpose styles under informative files names
* Restructure size variables to utilize `rem` over `px` (for zoom responsiveness)
* Update global styles to use font & colors from wireframes
* Adds consistent focus styling globally for accessibility

Structure
* Create `shared components/` to factor out repetitive sections of code 
* Create `models/` to hold applicable enums/interfaces/data structures
* Create `helpers/` to hold helper function files (to factor out cluttering/re-utilizable functions)

[See wireframes here](https://www.figma.com/file/TM86P6ePUar5g24pk4h6Du/NPDC-2?node-id=184%3A13)

## Screenshots:

ViewerRegistration
Captura de pantalla 2021-07-06 a las 10 07 52 p  m
Captura de pantalla 2021-07-06 a las 10 08 16 p  m
Captura de pantalla 2021-07-06 a las 10 10 40 p  m

UserLogin
Captura de pantalla 2021-07-06 a las 10 07 43 p  m
Captura de pantalla 2021-07-06 a las 10 10 47 p  m


srsexton94:
Updated to make pages mobile responsive & remove `reset.css` as per discussion yesterday

JMStudiosJoe:
57 files changed?

srsexton94:
@JMStudiosJoe yeah it's a chunky ticket - involved a lot of structural setup in the repo & establishing precedent re: typescript practices

IMHO the best review strategy would probably be to pull the branch & look at it in-editor (not a huge fan of github's PR UI 😅)

frontend/helpers/syntax-helper.tsx
```diff
@@ -0,0 +1,16 @@
+export default class SyntaxHelper {
```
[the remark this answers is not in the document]
JMStudiosJoe: ya can we use these more in css than javascript ?

frontend/pages/shared-components/enrollment/call-to-action/call-to-action.module.css
```diff
@@ -0,0 +1,11 @@
+.ctaContainer {
```
JMStudiosJoe, line 1: I think shared-components should go outside of pages directory. From what I read in the Nextjs docs all pages that are to be routable should go into pages directory, can we move the share-components to outside the pages directory?

frontend/pages/shared-components/enrollment/header/header.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react' 
+import styles from './header.module.css'
+import { InfoTooltip, Logo } from '../../index'
+import { TooltipTypes } from '../../../../models'
```
JMStudiosJoe, line 4: two things. 
1. having many directories creates a higher stack height in the directory tree. Can we get rid of the /enrollment directory and just have the components in the shared-components/ directory? 

2. Is this just the skeleton for the header component? Someone else took on the story for the header component so hopefully this wont step on too many toes.

JMStudiosJoe: disregard 2.

frontend/pages/shared-components/info-tooltip/info-tooltip.tsx
```diff
@@ -0,0 +1,32 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faQuestionCircle } from '@fortawesome/free-regular-svg-icons'
+import React from 'react' 
+import styles from './info-tooltip.module.css'
+import { tooltipContent, TooltipTypes } from '../../../models'
+
+interface InfoTooltipProps { type: TooltipTypes }
```
JMStudiosJoe, line 7: Can we put in a TODO here to factor out the tooltip container from the content somehow? I ask since the basic info tooltip is great but what if we have another more customized tooltip that displays more than basic info? instead of the `tooltipConent` import can we pass that as a prop from the parent and just display the `{ children }` that way we can reuse the base Tooltip. Does this make sense? can definitely do in another PR but should add at least a TODO.

frontend/pages/shared-components/text-input/text-input.module.css
```diff
@@ -0,0 +1,34 @@
+.inputContainer {
+  display: flex;
```
JMStudiosJoe, line 2: lets try to use css-grid instead of flex, far more clean and declarative.

frontend/pages/shared-components/text-input/text-input.tsx
```diff
@@ -0,0 +1,49 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons'
+import React, { FormEvent, useState } from 'react'
```
JMStudiosJoe, line 3: follow up of a previous comment to see these shared-components outside of the pages/ directory

frontend/pages/shared-components/text-input/text-input.tsx
```diff
@@ -0,0 +1,49 @@
...
+import styles from './text-input.module.css'
+import SyntaxHelper from '../../../helpers/syntax-helper'
```
JMStudiosJoe, line 5: Not too much of a fan on this since for me does not separate concerns between functionality of javascript and text UI manipulation which imo should only be through CSS

frontend/pages/user-login/user-login.module.css
```diff
@@ -0,0 +1,9 @@
+.login {
+  display: flex;
```
JMStudiosJoe, line 2: follow up to another comment, lets try to use css-grid over flex.

frontend/pages/user-login/user-login.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { FormEvent, useState } from 'react'
+import styles from './user-login.module.css'
+import { EnrollmentCTA, EnrollmentHeader, SubmitButton, TextInput } from '../../shared-components'
```
JMStudiosJoe, line 3: I think this component would be best inside of the compositons/ directory same with registration since the registration & login pages just import them

frontend/pages/viewer-registration/password-aid/password-aid.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { passwordToggleViews } from '../../../models'
+
+interface PasswordAidProps { id: string, onDisplayChange: any }
+export default function PasswordAid({ id, onDisplayChange }: PasswordAidProps) {  
+  const { passwordAid, passwordToggle } = styles
+  const { showView, hideView } = passwordToggleViews
+
+  const [{ icon, isHidden, text }, setPasswordView] = useState(showView)
+  
+  function handlePasswordView($event: MouseEvent<HTMLButtonElement>): void {
```
JMStudiosJoe, line 13: Also thought we were trying to avoid typing everything and were only gonig to type the same data from the backend? This is just my dislike of typescript needing to type everything where I have spent too much time debugging this rather than just getting actual work in.

frontend/pages/viewer-registration/viewer-registration.tsx
```diff
@@ -0,0 +1,50 @@
...
+    setSubmission(true)
+  }
+
+  return (
+    <section className={registration}>
+      <EnrollmentHeader
+        headerText="Register: Viewer Account" 
+        tooltip={TooltipTypes.VIEWER}
+      />
+      <form onSubmit={handleSubmit}>
```
JMStudiosJoe, line 31: please remove the form element just use a button with onClick.

JMStudiosJoe: will then reduce the fieldset use and simplify the component

frontend/shared-components/enrollment/call-to-action/call-to-action.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import Link from 'next/link'
+import styles from './call-to-action.module.css'
+import { CTATypes, enrollmentCTAText } from '../../../models'
```
JMStudiosJoe, line 4: this seems more like a sharable composition than a generaric component?

frontend/shared-components/info-tooltip/info-tooltip.module.css
```diff
@@ -0,0 +1,59 @@
+.tooltipContainer {
```
JMStudiosJoe, line 1: is this the same info-tooltip from pages?

frontend/shared-components/logo.tsx
```diff
@@ -0,0 +1,6 @@
+import React from "react"
+import LogoSVG from '../public/NPDCLogo.svg'
+
+export default function Logo() {
```
JMStudiosJoe, line 4: should likely have only one place with all logos.

frontend/shared-components/submit-button/submit-button.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react"
+import styles from './submit-button.module.css'
+
+interface SubmitButtonProps { text?: string }
+export default function SubmitButton({ text }: SubmitButtonProps) {
+  const displayText: string = text || 'Submit'
+
+  return (
+    <input
```
JMStudiosJoe, line 9: lets keep this as a simple button with an onclick using form and input submit has other side effects that can cause lots of issues down the road, better to simplify and just use a button.

frontend/shared-components/text-input/text-input.tsx
```diff
@@ -0,0 +1,48 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons'
+import React, { FormEvent, useState } from 'react'
+import styles from './text-input.module.css'
+import SyntaxHelper from '../../helpers/syntax-helper'
+import { InputNames, inputValidation } from '../../models' 
+
+interface TextInputProps { inputName: InputNames, isSubmitted: boolean, isPasswordShown?: boolean  }
```
JMStudiosJoe, line 8: duplicate component?

JMStudiosJoe (changes requested):
what a pr. Great work but many questions and comments.

frontend/helpers/syntax-helper.tsx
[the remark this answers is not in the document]
srsexton94: I'm not sure how one would use css for these functions, and the values they return are used programmatically, not just in browser display 

If you can specify a little more what you're suggesting I'm happy to give it a try though!

frontend/pages/shared-components/enrollment/call-to-action/call-to-action.module.css
srsexton94: It is outside of the pages directory! Are you perhaps looking at an old version? I've squashed all my commits so I'm not sure why this displayed like that 
Captura de pantalla 2021-07-08 a las 12 04 08 p  m

frontend/pages/shared-components/enrollment/header/header.tsx
srsexton94: Oh okay! Sounds good - I'll squash the enrollment subdirectory in my next push!

frontend/pages/shared-components/text-input/text-input.module.css
srsexton94: Flexbox & Grid serve somewhat different purposes! I've always heard the advice when you're dealing more with simple, single-dimensional layouts (like I feel like we are here) that flexbox is preferable to Grid
And especially in regards to mobile responsiveness, I find flexbox to be more dynamic? But happy to discuss further!

Here's an article I found helpful!
https://css-tricks.com/css-grid-replace-flexbox/

frontend/shared-components/enrollment/call-to-action/call-to-action.tsx
srsexton94: I'm not familiar with that distinction! Do you mind elaborating?

frontend/shared-components/submit-button/submit-button.tsx
srsexton94: Oh okay, sounds good! Will do

frontend/shared-components/text-input/text-input.tsx
srsexton94: No? the wrapping directory holds the tsx & css files together if that's what you're referring to?

frontend/shared-components/logo.tsx
srsexton94: I believe this is the only one so far - are you seeing somewhere else with logos?

frontend/shared-components/info-tooltip/info-tooltip.module.css
srsexton94: I think GitHub's PR UI for some reason might've been showing you an outdated structure (I had originally had `shared-components/` inside of `pages/` but moved it a level higher before pushing for PR)

I do believe there is only one tooltip so far!

frontend/pages/shared-components/info-tooltip/info-tooltip.tsx
srsexton94: Hey! Yeah I have a tooltip-related TODO in the related model's file - happy to add detail about this suggestion, but feels like it might be out of scope to tackle it in this ticket since ViewerRegistration doesn't utilize that (& as you pointed out, this ticket is already fairly large 😅)

frontend/pages/shared-components/enrollment/call-to-action/call-to-action.module.css
srsexton94: To clarify for other folks - turns out I introduced an accidental duplicate of the directory 🙈 will remove the one under `pages/` on next push!

frontend/pages/viewer-registration/password-aid/password-aid.tsx
srsexton94: I think we might need to override some typescript settings to not do that - it was throwing an error without a type

frontend/pages/viewer-registration/viewer-registration.tsx
srsexton94: Happy to use the button! but why no fieldset? I need a container of alike inputs & it's better to use a semantic element than a div or something

frontend/pages/user-login/user-login.tsx
srsexton94: Hmm? I'm sorry I'm not seeing a `compositions/` directory?

frontend/pages/shared-components/info-tooltip/info-tooltip.tsx
srsexton94: Oh, wait! This is already dynamically rendering content based on the "type" prop - so whether we select the content in the parent & pass it to the child, or pass the selector to select it in the child, seems rather inconsequential?

Personally, I like passing a selector (ie `type`) so that the parent has a concise label for the content, which makes reading the parent template code a bit clearer
Because of that, I'm inclined to keep it - but happy to keep discussing!

frontend/pages/shared-components/text-input/text-input.tsx
srsexton94: Resolving here to keep the conversation in the related thread above!

frontend/pages/viewer-registration/viewer-registration.tsx
srsexton94: Oh an as for the form element it's important for accessibility that things are semantically what they are performing as - if this is not a form element it would be inaccessible to screen reader users & fail any audit (WCAG 1.3.1)

frontend/pages/user-login/user-login.tsx
srsexton94: Resolving as I presume this was about the duplicate `shared-components/` directory (which has now been addressed)

If that was in error, happy to revisit this!

frontend/helpers/syntax-helper.tsx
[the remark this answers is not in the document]
JMStudiosJoe: These seem like extra methods for something? the CapitalizeFirstChar can be done with `text-transform: capitalize;` but what are these js string manipulating functions for?

dylanesque (changes requested):

frontend/models/password-aid.tsx
```diff
@@ -0,0 +1,14 @@
+
+import { IconProp } from '@fortawesome/fontawesome-svg-core'
```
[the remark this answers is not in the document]
JMStudiosJoe: I think I did get this issue but after re running yarn install issue was resolved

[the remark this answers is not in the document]
dylanesque: Works for me!

dylanesque (approved):

frontend/shared-components/text-input/text-input.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const { errorMessageText, pattern, inputType } = inputValidation[inputName]
+
+  const [inputId, errorId] = [`${inputName}Input`, `${inputName}Error`]
+  const labelText: string = `${SyntaxHelper.getTitleCaseFromCamel(inputName)}:`
+  const displayType: string = isPasswordShown ? 'text' : inputType
+
+  const checkIsValid = (value: string): boolean => { return !isSubmitted || pattern.test(value) }
+
+  const [inputValue, setInputValue] = useState('')
+  const [isValid, setValidity] = useState(checkIsValid(''))
```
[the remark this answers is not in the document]
srsexton94: Ahh okay, so that would be like `isValid` and `setIsValid`? I'll do that on next push!

frontend/pages/viewer-registration/password-aid/password-aid.tsx
```diff
@@ -0,0 +1,29 @@
...
+import styles from './password-aid.module.css'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { passwordToggleViews } from '../../../models'
+
+interface PasswordAidProps { id: string, onDisplayChange: any }
+export default function PasswordAid({ id, onDisplayChange }: PasswordAidProps) {  
+  const { passwordAid, passwordToggle } = styles
+  const { showView, hideView } = passwordToggleViews
+
+  const [{ icon, isHidden, text }, setPasswordView] = useState(showView)
```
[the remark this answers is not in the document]
srsexton94: Ooh okay, it's been a looooong time since I've used hooks but this'll be a good learning op - thanks!

frontend/helpers/syntax-helper.tsx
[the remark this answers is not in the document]
srsexton94: It's so we can pass a single prop of the camel case input name (ie `firstName`) maintain that for the different IDs we need but still display the input label as in the wireframes (ie `First name:`)

And since 1) we need those IDs programmatically 2) we want the displayed label to correlate to those IDs dynamically (so that the component is reusable) and 3) it's not just capitalizing/de-capitalizing the letters, but also manipulating spacing - I don't think we can use CSS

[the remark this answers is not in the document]
srsexton94: `getCamelCase` though actually is for an earlier rendition of the ticket & is no longer being used - I kept it around as I imagined it could be useful later, but I'm happy to delete that one if that helps address the concern!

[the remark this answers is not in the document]
srsexton94: You mean just typing the inputs/outputs? I don't think that's particularly complex - especially for simple datatypes like these - and I think typing adds a lot of value in error-prevention

But in either case - I think we'd need to override some TypeScript settings in order to *not* type them this way, as I was getting errors without typing

frontend/package.json
```diff
@@ -30,8 +36,11 @@
     "lint-staged": "^11.0.0",
     "prettier": "2.3.0",
     "react-test-renderer": "^17.0.2",
+    "sass": "^1.35.1",
+    "sass-loader": "^12.1.0",
     "ts-jest": "^27.0.3",
-    "typescript": "^4.2.4"
+    "typescript": "^4.2.4",
+    "webpack": "^5.41.1"
```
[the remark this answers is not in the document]
srsexton94: Honestly, at this point, I don't remember 😅 I'll investigate & get back to you!

frontend/helpers/syntax-helper.tsx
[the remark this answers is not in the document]
dylanesque: Interesting...let's have a short sidebar at the next convenient juncture, I'm curious about what those compiler complaints had an issue with. If the param types are either correctly inferred or explicitly declared and the return value (or lack thereof) is declared TS def should not be complaining about helper functions.

frontend/package.json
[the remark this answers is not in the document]
srsexton94: I removed it and nothing seems to have changed so I'll go with that!

frontend/helpers/syntax-helper.tsx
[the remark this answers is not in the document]
srsexton94: Okay, sidebar sounds good! I don't see it as much of a slowdown and personally would really like to see us typing function inputs & outputs - but getting rid of the strict option would probably be a good idea nonetheless!

I'm going to resolve this thread for now but let's chat over it with the team on Tuesday!

frontend/pages/viewer-registration/password-aid/password-aid.tsx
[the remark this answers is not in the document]
srsexton94: For other folks, talked it over with Mike & we're gonna leave it as is for now & potentially address this in a tech debt ticket

srsexton94:
Addressed feedback - unsure of how to mark requested changes as "addressed" on GH but I'd love a re-review if y'all have got time!

JMStudiosJoe (approved):

PR #87, opened 2021-07-20
Custom hook

dylanesque:
After carefully examing the original code and use case, I'm shelving the vast majority of this ticket for now, and will revisit it if and when legitimate reuse issues prompt a refactor in the future. Two changes:

- Changing an 'any' type to 'Function' 
- Moving the password-aid component out of the pages directory: only page components and their corresponding CSS files should be in this directory.


frontend/pages/viewer-registration/index.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { FormEvent, useState } from 'react'
 import styles from './viewer-registration.module.css'
-import PasswordAid from './password-aid/password-aid'
+import PasswordAid from '../../shared-components/password-aid/password-aid'
```
srsexton94, line 3: This was particularly kept out of shared components because it won't be used anywhere else in the app except on the registration page

Is there a particular reason we can't have children components in their parent's directory?

srsexton94 (commented):
Curious why we didn't go forward with the custom hook?

dylanesque:
> Curious why we didn't go forward with the custom hook?

After much more careful examination of the password-aid code, I feel like it would be fairly arbitrary at this point in time, but that will likely change if and when we have another need for toggle logic similar to the show/hide password logic.

frontend/pages/viewer-registration/index.tsx
srsexton94: Okay, hmm It's a pattern that I'm used to but that works to - in this case though wondering if we might be able to revisit the idea of having sub-directories in the shared-components directory? 
Seems like shared-components is going to get crowded quickly if it's really just all non-page components

dylanesque: I think having one sub-directory level deep in the shared-components folder is more than reasonable.

srsexton94 (dismissed):

srsexton94 (approved):

PR #89, opened 2021-07-20
Feature85/enrollment UI design update

srsexton94:
This ticket creates & adjusts shared components, global styles, & models to maintain consistency between the three "enrollment" pages (user login, viewer registration, and passport application) and their corresponding wireframes

Some liberties were taken with error message phrasing to get them to conveniently fit their space but we can continue playing with the styling if we want to revert those

Screenshots:
Login:
Login_onload
Login_errors

Registration:
Register_onload
Register_errors

Passport:
Passport_onload
Passport_errors


frontend/shared-components/enrollment-input/enrollment-input.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { FormEvent, useState } from 'react'
+import styles from './enrollment-input.module.css'
+import { FormLevelError } from '../index'
+import { getTitleCaseFromCamel } from '../../helpers/syntax-helper'
+import { EnrollmentInputNames, enrollmentValidation } from '../../models' 
+
+interface EnrollmentInputProps{inputName: EnrollmentInputNames, isSubmitted: boolean, isShown?: boolean, size?: string}
```
JMStudiosJoe, line 7: why an interface and not just a type?

frontend/shared-components/error-form-level/error-form-level.tsx
```diff
@@ -0,0 +1,16 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons'
+import React from 'react'
+import styles from './error-form-level.module.css'
+
+interface FormLevelErrorProps { errorId: string, errorMessage: string }
```
JMStudiosJoe, line 6: Can we use just type instead of Interface since that implys more generic reusability which we are not following that programming paradigm.

JMStudiosJoe (approved):

PR #94, opened 2021-07-22
Bugfix88 🐛 misc accessibility fixes

srsexton94:
This ticket will:
* Add metadata to HTML head for accessibility requirements (HTML lang & page title) & seo/social media (keywords, descrip, open graph tags)
* Fix form validation so that it responds appropriates to submit button
* Adjust the Logo component so that it will take alt text & not be presented as a group
* Adjust roles & aria attributes on form elements for screen reader pronunciation
* Change "State" form element from a select element to an input with a data list so that it's
     A) searchable 
     B) more easily made legible to screen readers

As a mac user, I only tested on VoiceOver - probably sufficient for now, but somewhere down the line if someone has a PC we should probably test out NVDA


JMStudiosJoe (dismissed):

frontend/shared-components/state-input/state-input.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { FormEvent, useEffect, useState } from "react"
+import styles from "./state-input.module.css"
+import { FormLevelError } from '../index'
+import { capitalizeFirstChar } from '../../helpers/syntax-helper'
+import { states } from '../../models'
+
+interface USAStateInputProps { isSubmitted: boolean }
+export default function USAStateInput({ isSubmitted }: USAStateInputProps) {
+  const {stateSelect, hasError} = styles
+  const [inputId, listId, errorId] = ['stateSelectInput', 'stateSelectList', 'stateSelectError']
```
dylanesque, line 10: Why are these variables being declared thusly? This is very confusing.

frontend/shared-components/response-textarea/response-textarea.tsx
```diff
@@ -12,13 +12,14 @@ export default function ResponseTextArea({ isSubmitted }: ResponseTextAreaProps)
...
-    if (isSubmitted) setIsValid(currCharCount >= charMin)
-    setCharCount(currCharCount)
+  
+  function handleTextareaChange({ target: { value }}: FormEvent<HTMLTextAreaElement>): void {
+    if (isSubmitted) setIsValid(value.length >= charMin)
+    setCharCount(value.length)
   }
+  
+
+  useEffect(() => { if (isSubmitted) setIsValid(charCount >= charMin) }, [isSubmitted])
```
dylanesque, line 22: This `useEffect` seems unnecessary. Did you run into some sort of race condition or timing issue without it?

dylanesque (changes requested):
The `useEffect` usage here doesn't quite make sense, at least at first glance.

frontend/shared-components/response-textarea/response-textarea.tsx
srsexton94: The useEffect is needed so that the validation is only run once the user tries to submit the form - we don't want error messaging to show up before they've completed

frontend/shared-components/state-input/state-input.tsx
srsexton94: Ah okay - I figured 3-1 line brevity would be worthwhile here and I didn't think it'd be that confusing (presumably any React devs would be pretty familiar with array deconstruction)

But I'm not married to the convention - happy to swap it out for regular 1:1 declarations! 

OR looking again at it there is a fair bit of repetition between the names, we could also DRY it up by writing it out `[inputId, listId, errorId] = ['Input', 'List', 'Error'].map(word => 'stateSelect' + word)

ALSO noticing that it probably shouldn't be "stateSelect" anymore as it's no longer a select element - I'll update that as well!

dylanesque: I wouldn't go as far as to say 'declare them the typical way': I think a short comment near the top of any component that's declaring vars using this convention would fully address my readability/comprehension concerns. The code works and is more succinct than the alternative, but it is somewhat unorthodox.

etilden: This array deconstruction is a fairly common construct-- any time you're using `const [x, setX] = useState(z)` you're using that construct!

srsexton94: Hmm so I'm pretty against the idea of using comments for something so inconsequential - I'm also not as keen on my other idea of mapping (think it complicates things unnecessarily) 

So I think the viable options are either keeping it as is or doing the 1:1 way - @dylanesque any preference?

dylanesque (commented):
> This array deconstruction is a fairly common construct-- any time you're using const [x, setX] = useState(z) you're using that construct!

Right, destructuring is common, but arbitrarily grouping primitive variables together like this isn't a common convention. This is the first time I've ever seen the right-hand side of a restructuring assignment be something other than the value pointing to an array or object, and I feel like this going against the common restructuring use case of "upacks complex data types so you have easier access to their properties/members.

If we had to choose, I would favor more standard variable declaration, even if it's slightly more verbose.

frontend/shared-components/response-textarea/response-textarea.tsx
srsexton94: But we're _not_ preventing validation after form submission(?) We're enabling it - we're preventing validation _before_ submission

The logic so far is basically 
1) the user lands on an empty form by default & can fill it out, we don't want to validate their inputs yet as they haven't clicked "Submit" (also in accordance with POLA)
2) the user clicks submit which changes `isSubmitted` to true
3) **if there are any errors** we want to stay on the form & display said errors (as navigating away from a form in error would actually violate POLA)
4) **if there _aren't_ any errors** we would then navigate away from the page, submitting the data, but this logic isn't included yet as we have nowhere to actually send said data

I've run through it to make sure it works! You could pull my branch & serve it up to see - or I can get a screen recording later on too if that's helpful!

srsexton94:
> Right, destructuring is common, but arbitrarily grouping primitive variables together like this isn't a common convention. This is the first time I've ever seen the right-hand side of a restructuring assignment be something other than the value pointing to an array or object, and I feel like this going against the common restructuring use case of "upacks complex data types so you have easier access to their properties/members.
> 
> If we had to choose, I would favor more standard variable declaration, even if it's slightly more verbose.

It wasn't arbitrary & I was just confirming your preference. Like I said before, I'm not married to it & happy to make the switch. I will once I can resolve some merge conflicts/fix some errors introduced by a recent PR merge.

Also, jsyk, that's not the only intended use case for restructuring? [7 Use Cases for Destructuring](https://javascript.plainenglish.io/7-use-cases-for-destructuring-in-javascript-4c8a73b28855)

frontend/shared-components/response-textarea/response-textarea.tsx
dylanesque: Ok, I see what you're going for. My apologies, I got some really bad news today and I'm clearly not in a headspace where I can review code with the required clarity. Withdrawing my objection to this.

dylanesque (dismissed):

frontend/compositions/state-input/state-input.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+interface USAStateInputProps {
+  isSubmitted: boolean
+}
+export default function USAStateInput({ isSubmitted }: USAStateInputProps) {
+  const {stateSelect, hasError} = styles
+  const [inputId, listId, errorId] = ['stateSelectInput', 'stateSelectList', 'stateSelectError']
+
+  const [inputValue, setInputValue] = useState('')
+  const [isValid, setIsValid] = useState(!isSubmitted)
```
etilden, line 15: Setting isValid to !isSubmitted seems a little counterintuitive to me-- I see that you have checkIsValid below which you're using to set the value of isValid, and you're only updating isValid if isSubmitted. I'm curious if you're anticipating that we'd ever initialize this component with isSubmitted=true? 
In general can you explain why you  did it this way?

frontend/compositions/state-input/state-input.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  function checkIsValid(value: string): boolean {
+    const matchingState = states.filter(({initials, name}) => {
+      return value.toUpperCase() === initials || capitalizeFirstChar(value) === name
```
etilden, line 19: I'm curious about this-- it looks like we're handling the cases where a state abbrev. is submitted (i.e. MA) or a full state name is submitted (i.e. Massachusetts)-- is there a reason why we don't know which value we're getting

etilden: Also I'm curious why we're using .filter() over .find() -- are we anticipating that there would be more than one potential result? If we used .find() we could just return !!matchingState rather than assessing an array length.

frontend/shared-components/error-form-level/error-form-level.tsx
```diff
@@ -1,7 +1,7 @@
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons'
 import React from 'react'
 import styles from './error-form-level.module.css'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons'
```
etilden, line 4: Curious why we moved these!

frontend/shared-components/state-input/state-input.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { FormEvent, useEffect, useState } from "react"
```
etilden, line 1: I'm curious why we have two state-input.tsx files, one in compositions and one in shared-components

srsexton94: Oof that was an oversight resolving merge conflicts - great catch! 
I think we want the one that's in shared-components as it will likely be used across different pages so I'll remove the compositions one

frontend/compositions/state-input/state-input.tsx
srsexton94: Re: the first Q - the user can search through the provided data list either by the abbreviation or by the state name - in the event that they just type the state name/abbreviation without actually clicking to select it from the menu I don't want it to throw an error if it's otherwise valid

Re: the second Q - no reason! I guess filter just came to mind first lol - happy to switch to `.find()` to simplify the return!

srsexton94: Re: the first Q - the user can search through the provided data list either by the abbreviation or by the state name - in the event that they just type the state name/abbreviation without actually clicking to select it from the menu I don't want it to throw an error if it's otherwise valid

Re: the second Q - no reason! I guess filter just came to mind first lol - happy to switch to `.find()` to simplify the return!

frontend/compositions/state-input/state-input.tsx
srsexton94: Hmm, I can't quite remember honestly - I remember it used to be set to state `checkIsValid('')` but I ran into some sort of issue with that. That issue could be outdated by now though honestly so I'll try again!

frontend/compositions/state-input/state-input.tsx
etilden: Makes sense!! Thanks for clarifying! 

I do have a slight preference for .find() if you happen to be pushing again!

frontend/compositions/state-input/state-input.tsx
etilden: To me it makes more senses to initialize isValid to "true", unless there's a particular reason not to!

frontend/compositions/state-input/state-input.tsx
srsexton94: Find implemented!

frontend/compositions/state-input/state-input.tsx
srsexton94: To clarify for other folks, addressed this with emma to use `checkIsValid()` calls onInit!

etilden (approved):

PR #97, opened 2021-07-24
some reorganizing to match the definition of directions more so, adde…

JMStudiosJoe:
…d layout component for wrapping all pages for easier navigation tests updated to match project structure


JMStudiosJoe:
mostly project restructure along with setup for better navigation in app

frontend/compositions/state-select/state-select.tsx
```diff
@@ -1,38 +1,41 @@
...
+import { FormLevelError } from "../../shared-components/index"
+import { states } from "../../models"
 
-interface USStateSelectProps { isSubmitted: boolean }
+interface USStateSelectProps {
+  isSubmitted: boolean
+}
 export default function USStateSelect({ isSubmitted }: USStateSelectProps) {
-  const [selectId, errorId] = ['stateSelect', 'stateSelectError']
+  const [selectId, errorId] = ["stateSelect", "stateSelectError"]
```
dylanesque, line 10: I don't understand this particular code style: if the variables are strings, why not just declare them as strings?

frontend/pages/login/index.tsx
```diff
@@ -1,6 +1,7 @@
-import React, { FormEvent, useState } from 'react'
-import { EnrollmentCallToAction, EnrollmentHeader, EnrollmentInput } from '../../shared-components'
-import { CallToActionTypes, EnrollmentInputNames } from '../../models'
+import React, { FormEvent, useState } from "react"
+import { EnrollmentCallToAction, EnrollmentHeader, EnrollmentInput } from "../../shared-components"
+import { CallToActionTypes, EnrollmentInputNames } from "../../models"
+import Layout from "../../shared-components/layout/Layout"
```
dylanesque, line 4: To clarify, the children I'm seeing get wrapped here are getting a unique Layout component that no other page on the site is getting?

dylanesque (changes requested):
This is largely fine, I just want clarity on the Layout component and where it's appearing (which may necessitate some changes) and the strengthened typing of props for the Layout component and I'm basically good.

frontend/shared-components/layout/layout.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react"
+import styles from "./layout.module.css"
+
+type Props = {
+  children?: JSX.Element | JSX.Element[]
```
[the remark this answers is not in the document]
JMStudiosJoe: awesome will update thanks!

frontend/compositions/state-select/state-select.tsx
JMStudiosJoe: ya, this is just kinda a slick way to do it that will allow for expanding on later but my only change was the quotes and that was due to the linting not me by choice. It does work so I didn't question it at the time but may @srsexton94 can elaborate on the declaration here?

frontend/pages/login/index.tsx
JMStudiosJoe: yes, so far the headers are different, there may not need to be a reason to wrap registration/login/passport but I felt it couldn't hurt to add for clarification that all pages a user has access to should be wrapped by Layout.

Issue #66, opened 2021-06-21
[FEATURE] create US Map component


acharliekelly:
US Map component for multiple pages

![us-map](https://user-images.githubusercontent.com/46508773/122806211-0ecb5280-d298-11eb-8d71-20c6670e2c41.png)

**Requirements**
* Recognizable image of US
* Show data richness by state, via background color
* Show locations of data-rich cities
* Map must be usable as input control for search locations

-----------
edit 8/10/21

**NEXT STEPS**
- create a key for the map (see figma for design, use d3)
    -  include:
        color gradient block scale
        symbol guide 
    - should be able to add other types of keys and additional data
    - should be reusable for other types of visualizations (not tied to map-type)
- style markers according to the figma

**UPCOMING** / blocked
- [on hold] create pop up element (see figma) waiting for design approval + content
- [on hold] make map mobile-ready
- [on hold] interface/import files for map-data
-----------

sashamaryl:
I'm picking this this one up thanks @acharliekelly

PR #99, opened 2021-07-25
Adds map

sashamaryl:
starts map  -- #66 
- using d3 
- **using temp/fake data** 
- supports zooming, zoom-to-state
- supports markers/marker layer
   - currently renders for selected state


dylanesque (changes requested):
Looks good so far, except d3 hasn't been added?

sashamaryl:
> Looks good so far, except d3 hasn't been added?

oops -- thanks @dylanesque !
just pushed the fix

frontend/compositions/map/useData.tsx
```diff
@@ -0,0 +1,46 @@
...
+      value === ""
+        ? ([] as Feature[])
+        : value === "all"
+        ? data.features
+        : data.features.filter((d: Feature) => d.properties[property] === value)
+
+    return {
+      features: filteredFeatures,
+      filter: filterProperties,
+      setFilterProperties,
```
srsexton94, line 43: sorry for another nitpick! Trailing comma

frontend/compositions/map/useData.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return useMemo(() => {
+    const { property, value } = filterProperties
+
+    const filteredFeatures =
+      value === ""
+        ? ([] as Feature[])
+        : value === "all"
+        ? data.features
+        : data.features.filter((d: Feature) => d.properties[property] === value)
```
srsexton94, line 38: Hmm I feel like this could be simplified as a single ternary (checking if value === 'all' and then returning all or the filtered array) 
And then we could either capture the `value === ''` edge case in the final return statement OR I'm wondering... would `data.features` ever have a feature whose `properties[property]` was an empty string? 
If not - the second part of your second ternary (ie currently line 38) would already capture that case! (ie return an empty array typed as Feature[])

frontend/compositions/map/map-helpers.tsx
```diff
@@ -0,0 +1,27 @@
+import * as d3 from "d3"
+
+
+
+
```
srsexton94, line 5: Curious why we have so much leading space between imports & file content/exports in some of these files?

I don't have a particular preference for either (having it or not) but I do want us to be consistent & don't see that convention elsewhere - maybe a question for the team but I think we should either _have_ or _not have_ the leading line space in all files

srsexton94 (changes requested):
Hey! Looking great - you have some super cool typing standards I was excited to see!

I mostly just reviewed for clean code concerns so far - thinking the comments discussion could be a useful one to have as a team tonight!

frontend/compositions/map/Map.tsx
```diff
@@ -0,0 +1,126 @@
...
+    <div
+      id="map-background"
+      ref={ref}
+      style={{
+        gridColumn: "20% 60% 20%",
+        height: "100vh",
+        position: "relative",
+      }}>
+      <div className="map-button">
+        <button className="primaryButton" id="zoom-reset">
```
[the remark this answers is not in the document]
sashamaryl: Yep, it resets the map to it's original position and scale. I grab this button as a d3 selection from inside the Map component. The action on it needs to be on a d3 selection and use a zoom behavior, which I haven't found a good way of passing down in a callback yet.

frontend/compositions/map/map-helpers.tsx
sashamaryl: …because I have reading/writing difficulties and adding extra space everywhere helps a ton. I think we can all have our preferences with some clever prettier settings...I'll look into it!

frontend/compositions/map/map-helpers.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return [(elementCenter[0] + viewportCenter[0]) / 2, (elementCenter[1] - viewportCenter[1]) / 2]
+}
+
+export function getExtent(el: Element) {
+  const { left, top, right, bottom } = el.getBoundingClientRect()
+  return [
+    [left, top],
+    [right, bottom],
+  ] as [[number, number], [number, number]]
```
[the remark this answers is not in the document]
sashamaryl: yes, this is specific to the way d3 describes bounds: a pair of coordinates, essentially a tuple of tuples I think.

frontend/compositions/map/useData.tsx
sashamaryl: Yeah, the temp data set was giving me some trouble filtering on an empty string, but ideally, yeah, we can simplify this like you say.

frontend/compositions/map/BaseMap.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+type BaseMapProps = {
+  projection: d3.GeoProjection
+}
+
+export default function BaseMap(props: BaseMapProps) {
+  const ref = useRef(null)
+  const { projection} = props
+
+  // gets geojson data holding borders for us, states, and counties with names
```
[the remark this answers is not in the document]
sashamaryl: These all (are intended to) fit in the 3rd category - clarifying obscure 3rd party things and external values. I do think D3 benefits from explanation--I'll keep working on the balance going forward.

frontend/compositions/map/Map.tsx
```diff
@@ -0,0 +1,126 @@
...
+export default function Map() {
+  const data = useData()
+  const ref = useRef<HTMLDivElement>()
+  const { width, height } = useResizeObserver({ ref })
+
+  const zoomRef = useRef<SVGSVGElement>()
+
+  const [markerParams, setMarkerParams] = useState<MarkerDescription[]>([])
+  const [shouldShowCities, setShouldShowCities] = useState<boolean>(false)
+  const [selectedState, setSelectedState] = useState<string | null>(null)
```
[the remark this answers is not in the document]
sashamaryl: Yeah, I like this way a lot--hopefully others are into it!

frontend/compositions/map/useData.tsx
sashamaryl: I thought trailing commas were in this year? xD

sashamaryl:
Thanks so much for your helpful comments @srsexton94!

frontend/compositions/map/BaseMap.tsx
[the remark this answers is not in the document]
srsexton94: Ahhh gotcha - so D3 requires us to name these variables this way? 

If there's any flexibility in naming I imagine we can capture a good chunk of it that way - but if it proves particularly cumbersome a few, succinct comments could certainly be a fair exception!

frontend/compositions/map/map-helpers.tsx
srsexton94: That's totally fair! I'm happy to have extra space if we use it consistently - prettier sounds like it could be a big help there - thanks for looking into it!

dylanesque:
@sashamaryl is this waiting to be merged? I don't want to add further review if there are big incoming changes coming

sashamaryl:
@dylanesque  Yeah, I'm just getting started with @srsexton94 's review. There will be changes before I'm ready to merge. Thanks!

frontend/compositions/map/BaseMap.tsx
```diff
@@ -0,0 +1,43 @@
...
+import * as topojson from "topojson"
+import { Topology } from "topojson-specification"
+
+type BaseMapProps = {
+  projection: d3.GeoProjection
+}
+
+export default function BaseMap(props: BaseMapProps) {
+  const ref = useRef(null)
+  const { projection} = props
```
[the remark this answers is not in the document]
dylanesque: Small issue, I would like to see a slightly more descriptive/unique name for the ref here.

frontend/compositions/map/Map.tsx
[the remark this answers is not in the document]
dylanesque: Very nice use of union types here!

dylanesque (dismissed):
Tentatively approving this, all I ask is that you give the ref for the Map a slightly better name.

dylanesque:
One other thing, that's not urgent: You'll want to change the import strategy at some point to import only the d3 modules you'll need, not the entire package, since it's delivering a massive 10kb over the wire..

JMStudiosJoe:
is this now ready for review/merging?

sashamaryl:
Yes--ready for review/merging. Thank you!

frontend/compositions/map/BaseMap.tsx
```diff
@@ -0,0 +1,46 @@
+import * as d3 from "d3"
+import { FeatureCollection } from "geojson"
+import React, { useEffect, useRef } from "react"
+import * as topojson from "topojson"
+import { Topology } from "topojson-specification"
+import styles from "./map.module.css"
+
+type BaseMapProps = {
```
dylanesque, line 8: We shouldn't be mixing type aliases into the codebase since we've decided on interfaces: This and any other type aliases in this ticket should be converted to interfaces.

frontend/compositions/map/BaseMap.tsx
```diff
@@ -0,0 +1,46 @@
...
+  projection: d3.GeoProjection
+}
+
+export default function BaseMap(props: BaseMapProps) {
+  const ref = useRef(null)
```
dylanesque, line 13: This can be improved with a more specific name for the ref

frontend/compositions/map/Map.tsx
```diff
@@ -0,0 +1,118 @@
+import * as d3 from "d3"
+import { Feature, Point } from "geojson"
+import { useEffect, useRef, useState } from "react"
+import useResizeObserver from "use-resize-observer"
+import BaseMap from "./BaseMap"
+import { MarkerDescription, MarkerLayer } from "./marker-layer"
+import useData from "./useData"
+import styles from "./map.module.css"
+
+export type BoundingType = [[number, number], [number, number]]
```
dylanesque, line 10: I think the intention of this declaration can be improved with [narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html): Could we try something like:

```
interface BoundingType {
xStart: number;
yStart: number;
xEnd: number;
yEnd: number;
}
```

frontend/compositions/map/Map.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  zoomContainer.on("click", onMapClick)
+
+  return (
+    <div
+      id="map-container"
+      className={styles.mapContainer}
+      ref={mapRef}
+      style={{ height: "80vh", width: "100vw" }}>
```
dylanesque, line 97: Small nitpick: I know that mobile use-cases are not our primary concern, but we should avoid using `vh` in most cases since that value gets unpredictable on mobile devices: https://css-tricks.com/the-trick-to-viewport-units-on-mobile/

frontend/compositions/map/useData.tsx
```diff
@@ -0,0 +1,46 @@
+import { Feature } from "geojson"
+import { useMemo, useState } from "react"
+import uscities from "./uscities.json"
+
```
dylanesque, line 4: Since this a custom hook, the 'hooks' folder would be the best place for it to live.

dylanesque (changes requested):
This is a LOT of work on a tricky feature, so seriously, well done. I've left comments about specific concerns I have, and I'd also suggest a 'map-data' files in the 'models' directory for all of the interfaces, along with the Angular-style import files that @srsexton94 has been utilizing to clean things up.

sashamaryl:
Thanks for your review @dylanesque !!
I've got a few answers/follow-up questions for you on some of the comments. Look forward to discussing with you -- maybe we could grab a breakout room briefly on Tuesday?

dylanesque:
> Thanks for your review @dylanesque !!
> I've got a few answers/follow-up questions for you on some of the comments. Look forward to discussing with you -- maybe we could grab a breakout room briefly on Tuesday?

I'll be out on Tuesday, but let's find a time soon (my early Saturday is pretty free this weekend) to hop on a call and get those questions addressed.

frontend/compositions/map/Map.tsx
sashamaryl: d3 requires this to be [[number, number], [number, number]] , and any bounds method will return this type, so we'll always encounter it.  It does make sense to have it this way: it is describing a bounding box using two points - upper left and lower right.

frontend/compositions/map/Map.tsx
sashamaryl: Thanks for flagging this and sharing the article (I LOVE css-tricks).
We need a absolute size here, and we need to set it directly on the div (not in the CSS), so the solutions in this article wouldn't work in this case.  
Since it's not primary, and the task might be a pretty big one, we should hold on this and focus on primary features. The way the map fits into the rest of the layout will be changing the container anyway, so we should at least wait to see what we end up with then.

frontend/compositions/map/Map.tsx
dylanesque: That's unfortunate, but requirements are requirements. I'd like to look at the docs and see if there's some way to make this more descriptive, but it's not a priority or a blocker.

frontend/compositions/map/Map.tsx
dylanesque: Yeah, again not urgent, but something for us to keep in mind since it's such a major gotcha with vh

frontend/compositions/map/useData.tsx
JMStudiosJoe: is this still applicable change to fix?

JMStudiosJoe (approved):
lgtm but im still sick @dylanesque @srsexton94 would you like to go through it again or ready to merge it?

PR #100, opened 2021-07-26
Data table

### LINKED ISSUE #90: [FEATURE] Results data grid component
After running a search, the results are displayed as a data grid, which will look something like this:

results-grid

Exact column names may change, but should have pagination and alternating colors for readability.

dylanesque:
Ready for review! The scope of this ticket is limited to the appearance of the table. I deliberately ignored the positioning of the table here, since we're missing interceding components that will make a big difference in exactly how it needs to be approached.

Fixes #90


JMStudiosJoe:
for some reason lots of `node_modules` was added

dylanesque:
> for some reason lots of `node_modules` was added

Yeah, that's annoying, I'll try and fix that

frontend/shared-components/data-table/data-table.tsx
```diff
@@ -0,0 +1,178 @@
...
+      },
+    ],
+    []
+  )
+  const tableInstance = useTable({ columns, data })
+
+  const { getTableProps, getTableBodyProps, headerGroups, rows, prepareRow } = tableInstance
+
+  return (
+    <div className={dTableWrapper}>
```
dylanesque, line 131: @srsexton94: Should I add a <caption> to this table, perhaps on that's visually hidden?

frontend/shared-components/data-table/data-table.tsx
```diff
@@ -0,0 +1,178 @@
...
+          ))}
+        </thead>
+        <tbody {...getTableBodyProps()}>
+          {rows.map((row) => {
+            prepareRow(row)
+            return (
+              <tr {...row.getRowProps()}>
+                {row.cells.map((cell) => {
+                  console.log(cell)
+                  if (cell.column.id === "full" || cell.column.id === "save") {
```
srsexton94, line 149: We could deconstruct the `id` directly from the parameter to avoid writing its reference over again!
Also, the lines 150-154 can probably appropriately be a one-liner! 

Generally, separation of concerns-wise we probably want to keep the amount of logic we perform directly in the template code pretty minimal - I've always gone by the rule of thumb that if it extends past 2-3 lines it should be factored out or simplified

frontend/styles/globals.css
```diff
@@ -38,3 +38,19 @@ a:visited {
...
+
+table {
+  border-collapse: collapse;
+}
+
+td,
+thead,
+tfoot {
+  font-family: Inter;
+  height: 36px;
```
srsexton94, line 50: can we add `--size36: 2.25rem` to `variables.css` and use it here!

Also, I would've thought `globals.css:19` would already cover the font-family - is that not the case?

frontend/styles/globals.css
```diff
@@ -38,3 +38,19 @@ a:visited {
...
+  text-align: center;
+}
+
+tr:nth-child(odd) {
```
srsexton94, line 54: If these styles are something specific to the table you're building (as opposed to something general to all tables we could want to build across the app) they should probably be scoped locally to your component - rather than in the global stylesheet

frontend/shared-components/data-table/data-table.tsx
```diff
@@ -0,0 +1,178 @@
...
+type IncidentData = {
+  dates: string
+  incidentType: string
+  officersInvolved: string[]
+  subject: string
+  source: string
+}
+
+export function DataTable({ count = 220375 }) {
+  // Defines table rows
```
srsexton94, line 18: This doesn't seem like a necessary comment, I'm also not sure what it's referring to (`count`?) 

I think this would be better served by just giving its referent a more informative variable name (maybe `numTableRows`?)

frontend/shared-components/data-table/data-table.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const { dTable, dHeader, dFooter, dPagContainer, dRowPage, dTableWrapper } = styles
+  const [rowsShown, setRowsShown] = React.useState(7)
+
+  // TODO: When this gets changed from mocking to fetching the data from an api call, append
+  // the full and save values dynamically
+  const data = React.useMemo(
+    () => [
+      {
```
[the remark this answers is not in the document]
dylanesque: That changes is forthcoming, particularly since i need a longer mock to test pagination.

frontend/shared-components/data-table/data-table.tsx
```diff
@@ -0,0 +1,178 @@
...
+        dates: "2003/01/01",
+        incidentType: "Use of force",
+        officersInvolved: ["John Smith"],
+        subject: "unknown",
+        source: "News Article",
+        full: faGreaterThan,
+        save: faPlusCircle,
```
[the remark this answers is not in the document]
dylanesque: This is strictly for mocking, we'll just dynamically append these to each search result when working with real data.

frontend/shared-components/data-table/data-table.tsx
```diff
@@ -0,0 +1,178 @@
...
+      },
+    ],
+    []
+  )
+
+  const columns = React.useMemo(
+    () => [
+      {
```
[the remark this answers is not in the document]
dylanesque: That is a necessary convention of the `react-table` library.

frontend/styles/globals.css
dylanesque: That actually led to some confusion for me: Are we using literally the same font-family for every element on the site?

srsexton94: Currently yes, it's what Ed had specified in our typography - if we don't like that we can certainly revisit it in a design convo - but that's what's in wireframes ATM

frontend/shared-components/data-table/data-table.tsx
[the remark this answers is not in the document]
srsexton94: Oh okay - so is this just a WIP PR? 

I'd just like to make sure we capture that before merging into main - but if you're still working through stuff that's totally valid too!

frontend/shared-components/data-table/data-table.tsx
dylanesque: I'm definitely open to suggestions as to how to clean this up, this is the implementation detail I'm unhappiest with here.

frontend/styles/globals.css
dylanesque: I'm going to need more time with this one: I've run into the first major snag with CSS modules, which is that they make using combination selectors in that paradigm effectively impossible.

frontend/shared-components/data-table/data-table.module.css
```diff
@@ -0,0 +1,27 @@
...
+  text-align: center;
+}
+
+.dataFooter,
+.dataHeader {
+  background-color: var(--lightBlue);
+}
+
+.dataRows:nth-child(odd) {
+  background-color: rgba(124, 174, 215, 0.5);
```
srsexton94, line 21: Hey! So sorry this was updated in the wireframes after you started this ticket, but do you mind updating this to `#B1CFE7` instead? We don't want the color to be transparent

frontend/shared-components/data-table/data-table.tsx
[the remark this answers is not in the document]
srsexton94: Oh? I'm curious why it would require a React component at all? I'm just talking about a simple array object variable you could plug in directly where we have the opening/closing square brackets lines 18-145

(Also so sorry I forgot to check back in on this so long -- got into design stuff and forgot to check back in with PR reviews)

frontend/models/data-table.tsx
```diff
@@ -0,0 +1,12 @@
+import { IconDefinition } from "@fortawesome/free-solid-svg-icons"
+
+export interface IncidentData {
+  dates: string
+  incidentType: string
+  officersInvolved: string[]
+  subject: IconDefinition
+  source: IconDefinition
```
srsexton94, line 8: I'm not sure if we want to update this/our mock data based on the new search flow/wireframes? 
(Totally not necessary if this is a first rendition/placeholder, but just noting for future!)

frontend/shared-components/data-table/data-table.tsx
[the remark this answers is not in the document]
dylanesque: Oh yeah, you're right, I just need to import the array I'm feeding into the useMemo calls 🤡

JMStudiosJoe:
do we just need to fix this merge conflict and will be good to go?

JMStudiosJoe (dismissed):

Issue #104, opened 2021-07-28
[FEATURE] Add User DB


DMalone87:
**Is your feature request related to a problem? Please describe.**
Currently, we're lacking a user object that can be used for authentication.

**Describe the solution you'd like**
We need to add the required DB objects to handle user authentication. A good solution will cover all of the different user types:
* A basic user who has only provided an email address and other optional profile information
* A Passport user who has provided some additional information and must be approved by an administrator
* A Contributor user who has permission to update and add records to the index
* An Admin user

PR #106, opened 2021-08-04
Add Authentication Routes and Utils

ryanmahan:
Addresses #2, adds routes for receiving a token from the api to include as an Authorization Bearer token on future requests. Also adds backend functionality for easily confirming and requiring the JWT on routes and blueprints.

Changes:
+ routes under /api/v1/auth for login, registering as a public user, testing you're authenticated
+ test for above routes
+ blueprint before_request function for requiring authentication in any route in a blueprint
+ decorator to verify JWT before route
+ DTOs for registering/logging in users (were these organized correctly?)


backend/database/models/user.py
```diff
@@ -44,8 +44,13 @@ def compile_ci_string(element, compiler, **kwargs):
         return base_visit
 
 
+class UserRole(enum.Enum):
```
alexjball, line 47: Darrell also listed contributor and admin roles in #104

backend/tests/test_auth.py
```diff
@@ -0,0 +1,75 @@
+import pytest
+from backend.database import User
+from backend.auth import user_manager
+from flask_jwt_extended import decode_token
+
```
alexjball, line 5: WDYT about a test case for the /test endpoint? That would illustrate how to send the access token in an API request.

backend/routes/auth.py
```diff
@@ -1,112 +1,100 @@
...
+from ..schemas import UserSchema
+from ..dto import RegisterUserDTO, LoginUserDTO
+from flask_pydantic import validate
 
 
-bp = Blueprint("auth", __name__)
+bp = Blueprint("auth", __name__, url_prefix="/api/v1/auth")
 
 
+# TODO: Place cookie on users browser with JWT token
```
alexjball, line 20: Consider using flask_login for browser sessions and JWT's for programmatic access. flask_login handles setting/refreshing cookies for you. If not, we should remove the flask_login stuff in user.py

backend/tests/test_auth.py
```diff
@@ -0,0 +1,75 @@
...
+)
+def test_login(client, db_session, password, expected_status_code):
+    test_user = User(
+        email="test2@email.com",
+        password=user_manager.hash_password("password"),
+        first_name="first",
+        last_name="last",
+    )
+    db_session.add(test_user)
+    db_session.commit()
```
dwreeves, line 38: You can create a fixture that creates a test user. The package `pytest-flask-sqlalchemy` rolls back the state of the db for each test so it's safe to continually recreate the same thing again and again.

Because we can guarantee the state of the database thanks to the aforementioned package and how we fire up the test database (reset it each test run), it also becomes much easier to test the case that an email _doesn't_ exist. So I would add `email` as a test parameter, and test for non-existing emails.

Finally, I would also use the string `"my_password"` instead of `"password"` because "my_password" should never occur anywhere in the code base other than in the tests. Protects against the possibility someone accidentally makes it so the code is checking for, literally, `"password"`.

The combination of all the above would be something like this. Haven't run this code but I think you can copy paste to replace from here to the end of the file

```python
@pytest.fixture
def example_user(db_session):
    user = User(
        email="test@email.com",
        password=user_manager.hash_password("my_password"),
        first_name="first",
        last_name="last",
    )
    db_session.add(user)
    db_session.commit()
    yield


@pytest.mark.parametrize(
    ("email", "password", "expected_status_code"),
    [
        ("test@email.com", "my_password", 200),
        ("test@email.com", "bad_password", 401),
        ("bad_email@email.com", "my_password", 401),
        ("test@email.com", None, 400)
    ],
)
def test_login(client, example_user, email, password, expected_status_code):
    res = client.post(
        "api/v1/auth/login",
        json={
            "email": email,
            "password": password,
        },
    )

    assert res.status_code == expected_status_code


 def test_jwt(client, example_user):
    res = client.post(
        "api/v1/auth/login",
        json={
            "email": "test@email.com",
            "password": "my_password",
        },
    )

    assert res.status_code == 200

    user_id = decode_token(res.json["access_token"])["sub"]
    db_user = db_session.query(User).filter(user_id == User.id).first()

    assert db_user.email == example_user.email
    assert res.status_code == 200
```

backend/tests/test_routes.py
```diff
@@ -16,14 +16,13 @@ def test_simple_routes(client, page, expected_status_code):
...
     )
     incident_id = res.json["id"]
 
-    incident_obj = db_session \
-        .query(Incident) \
-        .filter(Incident.id == incident_id) \
-        .first()
+    incident_obj = (
+        db_session.query(Incident).filter(Incident.id == incident_id).first()
+    )
```
dwreeves, line 25: Hrm. I might look into what Black is doing here. I do not agree with Black that this is more legible.

backend/api.py
```diff
@@ -3,18 +3,19 @@
...
 from .routes.incidents import bp as incidents_bp
-# from .routes.auth import bp as auth_bp
+from .routes.auth import bp as auth_bp
 from .utils import dev_only
 
 
 def create_app(config: Optional[str] = None):
     """Create the API application."""
     app = Flask(__name__)
+    app.config["JWT_SECRET_KEY"] = "super-secret"
```
dwreeves, line 18: Don't add this here! Add this to the `Config` object inside `config.py`, and set it like this. We don't want something like `key = "super-secret"` to ever touch our code, just in case we ever forget to swap it later.

```python
# ...
class Config(object):
    # ...
    JWT_SECRET_KEY = os.environ.get("JWT_SECRET_KEY", os.urandom(64))

**Review by dwreeves — CHANGES_REQUESTED** [review]
Really good PR! 🎉

The requested change to `api.py` is blocking.

The `example_user` fixture change is encouraged. If you agree with me that's better, then copy + paste in, otherwise let me know what you think.

I wouldn't worry about testing the state of the database for user registration. That's a good "first issue" for someone else to tackle as we onboard backend devs.

File: backend/tests/test_auth.py (hunk 4 above)
Comment (ryanmahan) [reply backend/tests/test_auth.py:38]: Getting an error from the DB implementing this, seems like the DB might not be resetting between tests?

```
sqlalchemy.exc.IntegrityError: (psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "user_email_key"
E       DETAIL:  Key (email)=(test@email.com) already exists.
```

File: backend/api.py (hunk 6 above)
Comment (ryanmahan) [reply backend/api.py:18]: Moved it, I kept the default as a `secret-key` instead of `urandom` so tokens could persist across local server restarts, is that ok?

File: backend/database/models/user.py (hunk 1 above)
Comment (ryanmahan) [reply backend/database/models/user.py:47]: Added!

File: backend/tests/test_auth.py (hunk 4 above)
Comment (ryanmahan) [reply backend/tests/test_auth.py:38]: Adding `db_session` as a parameter to the test that was causing this error fixed this problem

File: backend/tests/test_auth.py (hunk 2 above)
Comment (ryanmahan) [reply backend/tests/test_auth.py:5]: Added!

File: backend/routes/auth.py [hunk 7]
```diff
@@ -1,112 +1,100 @@
...
-        missing_fields = []
-        required_keys = ["email", "password"]
-        for key in required_keys:
-            if key not in form.keys() or form.get(key) is None:
-                missing_fields.append(key)
+@validate()
+def register(body: RegisterUserDTO):
+    # Check to see if user already exists
+    user = User.query.filter_by(email=body.email).first()
+    if user is not None and user.verify_password(body.password):
```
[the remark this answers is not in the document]
ryanmahan: Agreed, removed

backend/tests/test_auth.py
```diff
@@ -0,0 +1,75 @@
...
+@pytest.mark.parametrize(
+    ("email", "password", "expected_status_code"),
+    [
+        ("test@email.com", "password", 200),
+        ("bad_email", "bad_password", 400),
+        (None, None, 400),
+    ],
+)
+def test_register(client, email, password, expected_status_code):
+    res = client.post(
```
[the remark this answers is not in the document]
ryanmahan: Added a case for it

backend/routes/auth.py
ryanmahan: I can start working on that. I think we could make that a new issue and PR, so that this one doesn't get too much larger

backend/api.py
alexjball: That's okay for development but we shouldn't allow it in other environments. Maybe set the default to "secret-key" in development and urandom in other envs

alexjball (approved):
LGTM except for the default jwt secret! fine with me to merge after that's addressed

backend/tests/test_auth.py
dwreeves: excellent

Issue #49, opened 2021-06-16
[FEATURE] create Registration response page


acharliekelly:
Response screen for after user creates a new profile. Can use one page for all possible outcomes - only the message will change, depending on type of user and response (Viewer or Passport, success or failure).

**Success**
![profile-response](https://user-images.githubusercontent.com/46508773/122792367-1d117280-d288-11eb-8740-186728526420.png)

**Error**
![profile-response-error](https://user-images.githubusercontent.com/46508773/122792896-927d4300-d288-11eb-8072-62fcb68b2a90.png)

acharliekelly:
er, I don't seem to be able to add labels, but this should be labeled as frontend and good first issue

acharliekelly:
**Updated design, same requirements**
![profile-response](https://user-images.githubusercontent.com/46508773/122444577-b0883200-cf6e-11eb-8097-4b4a6b048151.png)

Devtinator:
Would like to work on this.

PR #108, opened 2021-08-05
Enrollment response

acharliekelly:
This is my first PR in a while, and also I don't know TypeScript, so there's a lot of stabbing in the dark. I'm trying to solve issue #49, which is theoretically a pretty good first issue. However, I do not understand the logic of breaking components up into tiny parts. Rather than asking someone to explain it to me, please comment on where things ought to have gone.

I created two components, RegistrationResponse and PassportApplicationResponse, which each take a single boolean `isSuccess` value to render one of two possible messages from the UI design. Unfortunately there isn't currently a way to see them in action, because they are just components. It should be easy enough to incorporate them into the Register and Passport forms, since all they need to know is whether the submission worked. But it seemed silly to modify either form without being able to do the thing that the response is signifying, so I left them alone.

I also created an external-link in shared-components, which I used for the "please alert our development team" link, and maybe we'll need another external link someday.


frontend/shared-components/external-link/external-link.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default function ExternalLink ({ linkPath, children }: ExternalLinkProps) {
+  return (
+    <a 
+      className={styles.externalLink} 
+      target="_blank"
+      rel="noopener noreferrer"
+      href={linkPath}
+      >
+        {children}
```
dylanesque, line 16: Is this intended to return children somewhere else not currently shown in this PR? I only see this wrapping a hardcoded string above here...

frontend/compositions/enrollment-response/enrollment-response.module.css
```diff
@@ -0,0 +1,29 @@
...
+  justify-content: left;
+}
+
+.message {
+  text-align: left;
+  line-height: var(--size24);
+  max-width: var(--size512);
+}
+
+.message p {
```
dylanesque, line 14: Are we certain that descendant selectors work as expected with CSS modules? I've had trouble with this working in my own tickets, so maybe take this review point with a grain of salt if you're certain that this is working as expected.

dylanesque (approved):
This looks good to me: I'd suggest destructuring the `styled` classNames across the board and clarify whether or not that `ExternalLinkPath` actually takes children, but I'm not seeing any blocking issues.

frontend/shared-components/external-link/external-link.tsx
acharliekelly: You're totally right, a string property would be fine

frontend/shared-components/external-link/external-link.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faExternalLinkAlt } from '@fortawesome/free-solid-svg-icons'
+
+interface ExternalLinkProps { linkPath: string, children: React.ReactNode }
+
+export default function ExternalLink ({ linkPath, children }: ExternalLinkProps) {
+  return (
+    <a 
+      className={styles.externalLink} 
+      target="_blank"
```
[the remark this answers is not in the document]
acharliekelly: That's a fair point. I guess the only point of the external link is to let the user know it's external.

acharliekelly:
Update: changed `ExternalLink` component so it doesn't open a new tab, and takes a string for link text. So really it's just a link. But it does have an external-link icon after it.

frontend/shared-components/external-link/external-link.tsx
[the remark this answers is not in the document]
dylanesque: If by the user you mean "other devs", that sort of thing is communicated in Next by the link not being wrapped in a <Link> tag, which is the sole domain of internal links.

acharliekelly:
Also, made snapshot test. Then, made snapshot test that successfully creates snapshot.

frontend/compositions/enrollment-response/enrollment-response.module.css
srsexton94: I've been using descendant selectors just fine if that helps!

frontend/compositions/enrollment-response/enrollment-response.tsx
```diff
@@ -0,0 +1,101 @@
...
+  returnText: string,
+  returnPath: string
+}
+
+export enum EnrollmentTypes {
+  VIEWER = 'viewer',
+  PASSPORT = 'passport'
+}
+
+const enrollmentMessage: { [key in EnrollmentTypes]: EnrollmentErrorText} = {
```
srsexton94, line 20: I think this & the above enums/interfaces (except for props) should probably live in our Models directory!

frontend/shared-components/external-link/external-link.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { faExternalLinkAlt } from '@fortawesome/free-solid-svg-icons'
+
+interface ExternalLinkProps { linkPath: string, linkText: string }
+
+export default function ExternalLink ({ linkPath, linkText }: ExternalLinkProps) {
+  const {externalLink} = styles
+  return (
+    <a 
+      className={externalLink} 
+      href={linkPath}
```
srsexton94, line 13: Oh awesome! Yeah so I think the screen reader warning just needs to be an aria-label so that screen reader will read linkText + "external" or something like that

[WebAIM on external links](https://webaim.org/techniques/hypertext/hypertext_links#new_window)

frontend/compositions/enrollment-response/enrollment-response.tsx
acharliekelly: That makes sense, but I don't understand where they should live, exactly. Is there some kind of method to how the Models directory is organized? I couldn't find any existing types that would naturally include "Viewer" & "Passport", even though they seem like obvious reusable parts of something

srsexton94: Hmm I'm not sure I totally follow - currently the models directory is just flat with a funnel `index.tsx` file to export everything in the directory (so that subsequent imports can be consolidated)

I'd totally be open to reorganizing that if that would help! I had just heard from others on the team that keeping a relatively flat structure was a priority for us

JMStudiosJoe:
@acharliekelly how goes it on this PR are you able to fix the confilicts and address the last of the comments?

acharliekelly:
Pretty sure conflicts are resolved, but still not 100% on git stuff

acharliekelly:
Oops, I think I must have broken one of the snapshot tests

JMStudiosJoe (approved):

PR #116, opened 2021-08-21
Set up CI for frontend

### LINKED ISSUE #115: [FEATURE] Run frontend Jest tests on Github Actions
Add a workflow to run frontend Jest tests.

alexjball:
Closes #115 

This creates a workflow that runs eslint, prettier, jest, and builds the production bundle. It runs all the checks every time we push or open a pull request under the frontend directory. If any of the steps fail, the "Frontend Checks" check will fail on the PR. Here is an example PR with the checks: https://github.com/alexjball/police-data-trust/pull/1

- Upgrade Next from 10 to 11 following [these](https://nextjs.org/docs/upgrading) instructions. This was necessary to use the integrated linting (easier than configuring separate linting)
- Use "strict" eslint rules following [these](https://nextjs.org/docs/basic-features/eslint#core-web-vitals) instructions
- Fix tests
- Fix eslint errors

Currently eslint warnings do not fail the check since there are several outstanding. We can make it stricter and fail on warnings if we want.
I have not fixed code formatting to keep this smaller. If this is merged I will follow up with a separate formatting PR.


JMStudiosJoe (approved):
lgtm besides the red checks do those need to pass before merging?

dylanesque:
How strict are these eslint rules going to be? I have concerns about overly strict airbnb style linting rules slowing stuff down....

alexjball:
@JMStudiosJoe The only failing check is the code formatting. I'll send out a followup with just the formatting changes. In general you're correct, should wait until checks are green before merging.

@dylanesque I don't think it's as strict as airbnb. You can read specifics [here](https://nextjs.org/docs/basic-features/eslint#eslint-config) but it enables react-related rules, next.js-related rules, and rules that affect "core vitals" (page load times). Seems like a reasonable place to start

PR #121, opened 2021-08-25
Data scraping

arutfield:
Currently will go to 4 sites, get all the data, and create the appropriate tables in excel


alexjball (changes requested):
Awesome work, Alex! Going to do another pass over the code but some requests after running the script:

- Put all files/folders created by the script in one `output` directory (either at the root or in scraper), and add the output directory to the `.gitignore`. This makes it easy to clean up after the script and identify all files it creates.
- I found it useful to have the raw data source downloaded for reference. The xlsx and csv handlers in `make_tables_data_source` directly convert the url to a dataframe so the original file isn't present after scraping. Consider downloading it like you do with the zip handler, writing it to a file, and loading the dataframe from that.
- Add the libraries you need to run the scraper to `requirements/_core.in` and run `flask pip-compile` (see readme in requirements folder). This lets you run the scraper after following the standard setup instructions 
- Rename the mapping_police_violence.py file to scrape_data_sources.py or similar, since it's not specific to that data source.
- Fill in the `scrape_command` in `backend/api.py`, so you can run the scraper with `flask scrape`:
```python
    @app.cli.command("scrape")
    def scrape_command():
        """Scrape from public data into the database.

        This is a handy way to populate the database to start with publicly
        available data.
        """
        from backend.scraper.mapping_police_violence import make_all_tables

        # TODO: load excel sheet into database
        make_all_tables()
```
- Update Readme to reflect above

backend/scraper/mapping_police_violence.py
```diff
@@ -162,12 +167,26 @@ def make_all_tables():
...
+    ])
+    final_indices = df_victim_filtered.index
+
     writer = pd.ExcelWriter('full_database.xlsx', engine='xlsxwriter')
 
     # Write each dataframe to a different worksheet.
     for key in table_dict:
         sub_data = table_dict[key]
-        sub_data.to_excel(writer, sheet_name=key)
+        sub_data.iloc[final_indices].to_excel(writer, sheet_name=key)
```
alexjball, line 189: I think this should use `.loc` rather than `.iloc`, since the index is a list of row labels, while `iloc` refers to the integer location of the row. I think it might work in this case due to our choice of index but that could change in the future.

backend/scraper/mapping_police_violence.py
```diff
@@ -63,12 +64,14 @@ def make_single_table(table, dat, configs):
...
     return dat[dat.columns[dat.columns.isin(cols)]]
 
 
 def extract_all_subfolders(head_directory, data_source, xwalk, dat, configs):
     if os.path.isdir(head_directory):
         for filename in os.listdir(head_directory):
+            print(filename)
             if configs["sources"][data_source]["subdirectory"] not in filename\
                     and configs["sources"][data_source]["subdirectory"]\
                     != "None":
```
alexjball, line 77: Below, you have a magic `nrows=1048576` in two places. Why do we need to specify that? It would be applied to all data sources, but I'd guess diff sources have diff numbers of rows.

backend/scraper/mapping_police_violence.py
```diff
@@ -128,6 +131,7 @@ def make_tables_data_source(data_source, xwalk, configs):
     dat.to_csv(data_source + '.csv', index=True, header=True)
```
alexjball, line 131: nit: move `dat = map_varnames... dat = gen_ids...` out of the elif clauses since it's the same in each

backend/scraper/mapping_police_violence.py
```diff
@@ -162,12 +167,26 @@ def make_all_tables():
         sub_table = full_df
         table_list.append(sub_table)
     table_dict = {table_names[i]: table_list[i] for i in range(len(table_list))}
+    df_victim = table_dict["victim"]
```
alexjball, line 170: So this is filtering the rows by removing rows with duplicate victim names, while retaining all rows with unknown victim names? 

- This would be a great thing to move into a separate deduplicate_rows function to hide these implementation details and express the intent more clearly. It would take the table_dict and update each dataframe with the filtered index. 
- I would just inline the victim_indices_* and rename the filtered_indices on line 174 to `valid_full_names`. 
- Consider renaming `df_victim_filtered` to something more descriptive like `deduplicated_by_name`
- You can use `~` to negate `valid_full_names` and save some typing: `deduplicated_by_name = np.concat([df_victim[~valid_full_names], df_victim[valid_full_names].drop_duplicates(...)])`

alexjball (changes requested):
Looking great! Happy to discuss more tomorrow. Also I haven't used pandas much so the related advice might be wrong :grimacing:

JMStudiosJoe:
@arutfield this is looking great is it ready for another review and to be merged?

backend/scraper/mapping_police_violence.py
```diff
@@ -147,6 +151,7 @@ def make_all_tables():
     data_list = [
         make_tables_data_source(x, xwalk, configs) for x in data_sources
     ]
+
     table_names = list(configs["tables"].keys())
```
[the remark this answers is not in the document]
arutfield: Not the same. table_names are the report, incident, victim, death. data_sources are the organizations we get the data from.

arutfield:
> @arutfield this is looking great is it ready for another review and to be merged?

Thanks! But not yet. I moved it back into draft to avoid confusion.

backend/scraper/mapping_police_violence.py
arutfield: We give it the maximum number of rows to ensure it has enough.

PR #145, opened 2021-09-07
Header updates

nancyyang94:
1. Updated mobile view 
- enlarged logo size
- removed donate button and title container 

2. Refactored `desktop-nav`
- followed first example from [https://flaviocopes.com/nextjs-active-link/] thanks @dylanesque for sharing this earlier), to use next.js router and `Link`


alexjball (changes requested):
Looks great! What will go in the mobile hamburger menu? Also, On mobile the blue header doesn't extend across the top

localhost_3000_dashboard

frontend/compositions/dashboard-header/desktop-nav.tsx
```diff
@@ -1,29 +1,32 @@
...
+  const router = useRouter()
 
-  const handleNavChange = (e: React.MouseEvent<Element>): void => {
-    const value = (e.target as HTMLElement).innerText
-    setSelected(value)
-  }
+  const menu = [
+    { title: "Search", path: "/dashboard" },
+    { title: "Profile", path: "dashboard/profile" },
+    { title: "Sign Out", path: "#" }
```
alexjball, line 12: We can create a /logout page that handles logging out and then redirects to the landing page. so here let's use `path: "/logout"`.

frontend/compositions/dashboard-header/desktop-nav.tsx
```diff
@@ -1,29 +1,32 @@
...
-  const [selected, setSelected] = React.useState("Search")
+  const router = useRouter()
 
-  const handleNavChange = (e: React.MouseEvent<Element>): void => {
-    const value = (e.target as HTMLElement).innerText
-    setSelected(value)
-  }
+  const menu = [
+    { title: "Search", path: "/dashboard" },
+    { title: "Profile", path: "dashboard/profile" },
```
alexjball, line 11: Charlie is using "/profile" in #150, so we should update this to match. It will also simplify where the page `tsx` lives

frontend/compositions/dashboard-header/desktop-nav.tsx
```diff
@@ -1,29 +1,32 @@
...
 export default function DesktopNav() {
-  const [selected, setSelected] = React.useState("Search")
+  const router = useRouter()
 
-  const handleNavChange = (e: React.MouseEvent<Element>): void => {
-    const value = (e.target as HTMLElement).innerText
-    setSelected(value)
-  }
+  const menu = [
+    { title: "Search", path: "/dashboard" },
```
alexjball, line 10: I think we should change the search route to "/search" for clarity. /search and /profile will share the header component. We will also need to update existing references to "/dashboard" to "/search". This could be done in a separate PR.

alexjball (commented):
If you want to avoid re-rendering the dashboard header, you could move it into the Layout component, and then use the Layout component from the main _app.js file rather than individual page comps. When a route changes, next re-renders MyApp with the new page component, so the header will change between routes.

```javascript
// pages/_app.js
<Layout/>
  <Component {...pageProps}/>
</Layout>
```

alexjball (changes requested):
Also, one of the tests is currently failing because the router isn't created in the test environment. You can fix that by installing `next-router-mock` and mocking the router:

`npm i -D next-router-mock`, then add this line to `jest.setup.ts`:
```
jest.mock("next/dist/client/router", () => require("next-router-mock"))
```

frontend/compositions/dashboard-header/desktop-nav.tsx
nancyyang94: @alexjball Hi Alex, thanks so much for the code review. I _think_ I’m understanding the main reason behind the suggestion to move Dashboard Header into <Layout>, but wanted to check in before I start shifting things around. 

My current understanding:
- Next’s Layout component is useful for holding components that are reused across every page, like a header, nav bar, or footer. This prevents unnecessary re-rendering of repeated components.
- We’ve setup a shared Layout component, but it’s currently not holding any components. 
- We’ve wrapped every page in the above Layout component. Since we don’t have anything in Layout currently, it’s behaving more like a wrapper than an actual Layout component.

Proposed changes:
- Move the Dashboard Header component into the Layout component. 
- Wrap MyApp(pages/_app.js) with the Layout component, and remove the Layout component from every page.
- Now, whenever a route changes, Next re-renders MyApp with the new page component (but avoids re-rendering the header, since it’s part of the set Layout now)

My question: 
 - Testing the proposed changes above locally, every page will now display the Dashboard Header at the top. However we don’t want to display the header for some pages (log-in, passport). Would a conditional check/display for the header, based on the current pathname, in Layout be enough to address this? I’m not sure if I’m overthinking this. 

Please let me know if I'm off-track in any of the above, or can provide any more context/details. 


- [ ]

frontend/compositions/dashboard-header/desktop-nav.tsx
nancyyang94: Will update.

nancyyang94:
> Also, one of the tests is currently failing because the router isn't created in the test environment. You can fix that by installing `next-router-mock` and mocking the router:
> 
> `npm i -D next-router-mock`, then add this line to `jest.setup.ts`:
> 
> ```
> jest.mock("next/dist/client/router", () => require("next-router-mock"))
> ```

Thank you - will do.

frontend/compositions/dashboard-header/desktop-nav.tsx
alexjball: Hey @nancyyang94 , great point, you are correct! putting the layout in _app would give every page the header, which we don't want to do for auth screens. the conditional approach could work. Overall I wonder if avoiding the rerender is worth the effort here, since it also makes things more complicated. I would lean toward keeping things as-is and revisiting if it becomes an issue.

nancyyang94: sounds good - will plan to keep things as is, and return to this if needed!

nancyyang94:
9/14:
Updated: 
- fixed failing snapshot test (installed next-router-mock) 
- addressed dashboard pathname update in separate PR #154. i made these changes in separate branch off main, `search-path-update` that's ready to be reviewed and merged if ready. 

WIP/to do this week -
- need to revisit mobile view bug 
- need to add mobile nav bar view

alexjball:
Thanks Nancy! merged the route change, so we can update the route here as well

acharliekelly (approved):
I agree with calling the route `/search` instead of `/dashboard`
Otherwise no issues

alexjball (dismissed):
Please update the snapshots, but otherwise LGTM!

nancyyang94:
> Please update the snapshots, but otherwise LGTM!

Sounds good - will work on the desktop view fix and update snapshots accordingly!

dylanesque:
> > Please update the snapshots, but otherwise LGTM!
> 
> Sounds good - will work on the desktop view fix and update snapshots accordingly!

Snapshots shouldn't need to be manually updated, there's a script for that in `package.json`:

```"test:update-snapshots": "jest --updateSnapshot"```

alexjball (approved):

Issue #141, opened 2021-09-02
[FEATURE] create Saved Results component


acharliekelly:
![profile_saved-searches](https://user-images.githubusercontent.com/46508773/131806127-3c3dacf3-5c2d-4d63-951c-e2df781235c4.png)

Display user's Saved Results. Should be able to edit collection and view each record.

acharliekelly:
included in #151

Issue #140, opened 2021-09-02
[FEATURE] create Saved Searches component


acharliekelly:
![profile_saved-results](https://user-images.githubusercontent.com/46508773/131805458-b9fbdd5a-6f3c-4f31-adfb-a82b944cf931.png)

Display user's Saved Searches in a data table. Should be able to edit the search collection and view the results of each search.

acharliekelly:
included in #151

Issue #139, opened 2021-09-02
[FEATURE] create Profile Type component


acharliekelly:
Show whether current user has regular or passport account.

![profile_type](https://user-images.githubusercontent.com/46508773/131805009-f9c26cfb-c80a-4d11-bd34-e7d05aa75e14.png)

**Requirements**
- Regular users should see link to register for Passport account. 
- Passport users should have the option to revert to a regular account.

acharliekelly:
included in #151

Issue #138, opened 2021-09-02
[FEATURE] create User Information component


acharliekelly:
User Info component for Profile page (default view of page). 
Should have an edit button so info can be changed and saved.

![profile_user-info_edit](https://user-images.githubusercontent.com/46508773/131804583-4c1a3648-a68b-415f-936c-17c3984d7517.png)

acharliekelly:
included in #151

PR #146, opened 2021-09-07
Profile page

acharliekelly:
Addresses #137, as well as #138, #139, #140, and #141, but none are what you'd call "finished." All the components need CSS work, and the Saved Search/Results components are only stubs, just so something happens when you click the menu items. Same with About page (`pages/about.tsx`), which I added for navigation completeness, but it has no content and should probably be replaced with `/pages/about/index.tsx`. Still in process of incorporating mock search data into saved-search components, but figured why not submit it now and get feedback.

Anyway, **please** make suggestions.


acharliekelly:
Ah, ok, need to fix tests, and also, apparently the header navigation has changed a bit. Will re-submit shortly

PR #150, opened 2021-09-08
Enrollment response

acharliekelly:
OK, let's try this again. Addresses #49,  again. Maybe.


frontend/models/response.tsx
```diff
@@ -0,0 +1,25 @@
...
+  returnText: string
+  returnPath: string
+}
+
+export enum EnrollmentTypes {
+  VIEWER = "viewer",
+  PASSPORT = "passport"
+}
+
+export const enrollmentMessage: { [key in EnrollmentTypes]: EnrollmentErrorText } = {
```
alexjball, line 14: For most literal values, you don't need to specify the type as the compiler can infer it. but this does prevent people from adding unrelated keys.

frontend/shared-components/external-link/external-link.tsx
```diff
@@ -0,0 +1,19 @@
...
+  linkPath: string
+  linkText: string
+}
+
+export default function ExternalLink({ linkPath, linkText }: ExternalLinkProps) {
+  const { externalLink } = styles
+  return (
+    <a aria-label={`${linkText} external`} className={externalLink} href={linkPath}>
+      {linkText}
+      <FontAwesomeIcon aria-hidden="true" icon={faExternalLinkAlt} size="xs" />
```
alexjball, line 16: nit: I think this icon implies the link opens in a new tab. Since this is a web app i do think it makes sense for external links to open in new tabs. Can update the link attributes with `target="_blank" rel="noopener noreferrer"`

alexjball (changes requested):
Pixel Perfect!

Until these are used in routes, could you create /passport/response.tsx and /register/response.tsx routes that just render these components in their two states?

```typescript
export default function Response() {
  return (
    <div>
      <PassportApplicationResponse isSuccess={true} />
      <PassportApplicationResponse isSuccess={false} />
    </div>
  )
}
```

frontend/shared-components/external-link/external-link.tsx
acharliekelly: That was in my first submission, but apparently forcing new tabs are controversial. Unfortunately my git problems prevented that discussion from appearing here, but Opinions were Expressed, so I took it out. We probably need a floor debate on the issue.

alexjball: alright--my understanding is that stateless pages should not open in new tabs but it's ok for web apps with state to open links in new tabs to avoid losing a user's progress. 🤷

acharliekelly: that works

alexjball (dismissed):
LGTM, need to update snapshots and fix formatting

JMStudiosJoe:
great work @acharliekelly lets update the formatting and snapshots and get this merged :)

Issue #137, opened 2021-09-02
[FEATURE] create Profile page


acharliekelly:
![profile_user-info](https://user-images.githubusercontent.com/46508773/131803260-e957322e-31c7-4925-ae79-6598d6c94850.png)

**Requirements**
Left menu with 4 items:
- User Information
- Profile Type
- Saved Results
- Saved Searches

Right side displays content

acharliekelly:
Fixed in #151

PR #151, opened 2021-09-08
Profile page

acharliekelly:
Addresses #137 (and also #138, #139, #140, and #141). Mostly works, except for:
- CSS is all wrong
- Code is not dry. Actually dripping wet. Suggestions appreciated.


frontend/compositions/profile-content/profile-info.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  function handleSubmit(e: React.FormEvent) {
+    // TODO: update user
+  }
+
+  if (editMode) {
+    return (
+      <div className={profileData}>
+        <header className={sectionTitle}>Edit Your Account Information</header>
+        <form onSubmit={handleSubmit}>
```
JMStudiosJoe, line 20: lets remove the onSubmit on the form, form has sideeffects when doing a POST request, place the `handleSubmit` on a normal button instead please

frontend/compositions/profile-content/profile-info.tsx
```diff
@@ -0,0 +1,80 @@
...
+            <label htmlFor="newPassword">Password:</label>
+            <input type="password" id="newPassword" />
+            <label htmlFor="confirmPassword">Confirm New Password</label>
+            <input type="password" id="confirmPassword" />
+          </div>
+          <div className={row}>
+            <button className={styles.cancelButton} onClick={() => setEditMode(false)}>
+              Cancel
+            </button>
+            <input type="submit" className={styles.submitButton} value="Save Changes" />
```
JMStudiosJoe, line 43: lets make this a regular button calling the `handleSubmit`

frontend/pages/profile/profile.module.css
```diff
@@ -0,0 +1,15 @@
+.profileWrapper {
+  width: 100%;
+  /* height: var(--size112); */
```
JMStudiosJoe, line 3: can remove

JMStudiosJoe (changes requested):
fantastic work so far just a few minor comments!

frontend/models/mock-search-meta.tsx
```diff
@@ -0,0 +1,63 @@
+import { faCaretRight, faAngleRight, IconDefinition } from "@fortawesome/free-solid-svg-icons"
```
alexjball, line 1: Is there a reason that mock-search-meta isn't in `__mocks__`? Also, `__mocks__` is used by Jest to mock modules, rather than to hold mock data. So I would suggest renaming `__mocks__` to mockData, moving all the mock data files there, and then having an `index.ts` file that re-exports everything for easier import.

frontend/pages/profile/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export default function Profile() {
+  const [nav, setNav] = React.useState(ProfileMenu.USER_INFO)
+
+  const mockUser = users[0]
+
+  return (
+    <Layout>
+      <DashboardHeader />
+      <div className={styles.profileWrapper}>
```
alexjball, line 26: Nice! You might also consider supporting nested routes like /profile/info, /profile/saved-results etc. You'd [create a file like [[...path]].tsx](https://nextjs.org/docs/routing/dynamic-routes#optional-catch-all-routes) and re-export this component. Then useRouter will give you the path segments.

alexjball (changes requested):
Looks great! The only change I'd really like to see is moving the mock data files.

We could be more DRY by using the PrimaryInput and DataTable components here. This is consistent with the figma mocks, which reuse the input/table components. The table component is styled very differently for search results and the profile screens but I wonder if we could share most other things there.

#152 introduces react-hook-form which automates some of the form control flow. I am repeating myself in there, so we could also try packaging the common form stuff into a shared hook.

frontend/models/mock-search-meta.tsx
acharliekelly: I wanted to make something that would be easy to switch over to using API data once that's working, but also work right away with Jest. However, a drawback of that approach is that I don't really understand how mocks work. Also, we could just put it in `models/` and then change one line of code when there's another source. But someday it would be good to have a mock data source for testing, if somebody knows how to do that.

frontend/pages/profile/index.tsx
acharliekelly: thanks!

frontend/models/mock-search-meta.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const resultsColumns = [
+  {
+    Header: "Search Date",
+    accessor: "searchDate" as const
```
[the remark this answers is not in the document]
acharliekelly: Would love to reuse table components - my saved search/results are made from copy-pasting the code from DataTable and then changing couple things. Didn't want to get my PR too far out of scope by fiddling with DataTable, but hoping somebody who's more comfortable with react-table would make something more generalized. 
Completely forgot about PrimaryInput, though!

frontend/models/mock-search-meta.tsx
acharliekelly: The mock-search-meta is a separate file in case the column names aren't available from the API

alexjball: Got it, I don't think we'll get column names from the API so consider renaming this to just search-meta.

I'm planning on adding [mock-service-worker](https://mswjs.io/) in #152 which will let us mock the API endpoints directly, and we can pipe in the mock data there. We can just leave the `__mocks__` as is as we'll need to rejigger things for that.

frontend/models/mock-search-meta.tsx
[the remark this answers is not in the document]
acharliekelly: Oh hey, let's merge #152 first - there's a bunch of stuff I can use in there, plus it seems dumb to muck around with fake users when real ones (or at least fake users via API instead of mocks) are so close

acharliekelly:
Updates:
- Improved CSS - looks more like FIGMA than before, but still needs work (also not responsive at all)
- DRYer tables
- moved mock data so it won't conflict with Jest mocks

alexjball (dismissed):
Please run `npm run fix` to fix the formatting and fix the broken test, but otherwise LGTM!
For the test, I think you need to wrap linkText in an `<a>` in profile-type.tsx. Check out the other usages of Link

alexjball (approved):
made some changes to pass CI, looks great! merging

PR #152, opened 2021-09-13
[frontend] Wire up auth flows

### LINKED ISSUE #109: [FEATURE] Make login form work
There is (or will be, pending #106) an API route for user authentication. But the login form was created by #48, which is closed; so there's no open issue for making the login form actually perform the authentication. So, that's what this is.

### Requirements
Modify the login form's `handleSubmit` function (in `pages/login/index.tsx` ) so that it:
- Sends the login fields to the API
- Uses the response to render either the dashboard or a login failure message

alexjball:
Closes #109 

New behaviors:
- login, register, and passport forms have validation and submission flows
- login and register forms are now connected to the backend API. The site is unusable without a running backend.
- /login and /register routes redirect to /dashboard if the user is authenticated
- /dashboard and /passport routes redirect to /login if the user is not authenticated, logs out, or their access token expires
- `npm run dev` by default mocks out the backend. Users can log in with the "test@example.com"/"password"

Other Changes:
- Create an initial test user in then backend development environment. email `test@example.com`, password `password`
- Set up CORS to allow all origins. This is necessary at least for the tests.
- Use [react-hook-form](https://react-hook-form.com/) for form input management.
- Simplify some of the input validation patterns as I couldn't figure out the required format and the error messages didn't clarify
- Create an AuthProvider to expose auth state/methods to components. auth info is stored in localStorage.
- Add a vscode workspace primarily used for debugging the frontend.


frontend/helpers/api/api.ts
```diff
@@ -0,0 +1,83 @@
...
+export type AccessToken = string
+
+export interface User {
+  active: boolean
+  email: string
+  emailConfirmedAt?: string
+  firstName?: string
+  lastName?: string
+}
+
```
acharliekelly, line 12: How do we know a User's role (passport, etc)?

acharliekelly (approved):
Wow, this fixes a lot more than just making the login work! 
Only issue I have is, the User interface (`frontend/helpers/api/api.ts`) doesn't have any role property, so how do we know whether they have a Passport account?

alexjball:
Thanks for the review @acharliekelly, and great question! We don't currently include the role in the User response. Created #155 to add it.

We haven't yet built or specified the passport flow. Who approves, and how? Admins in an admin interface? What happens if a user is rejected? Can they re-apply? We can discuss Tuesday. Adding role-related functionality including the passport flow could be its own milestone.

JMStudiosJoe (approved):

PR #157, opened 2021-09-21
update frontend docs

alexjball:


JMStudiosJoe (approved):
awesmoe thanks for doing this

Issue #76, opened 2021-06-30
[FEATURE] Add Storybook


acharliekelly:
Since we're going to be making reusable components, Storybook is a good way to test them in isolation.

This feature is just adding the dev dependency (and making sure it still installs). Actually making stories (tests) for components will be a separate thing.

alexjball:
Hey @acharliekelly what is the status on this? We discussed at the huddle today.

PR #170, opened 2021-10-01
Feat storybook

acharliekelly:
This PR incorporates Storybook into the project, per #76. Despite adding a number of files and packages, it has **No Production Footprint**. Currently displays all components in the `/shared-components`, `/compositions`, and `/pages` folders. There's a helpful USAGE.md file in the .storybook config folder. With the exception of two inexplicable pages, all components are rendering OK.

Side note, I couldn't get Sass to work (some kind of version problem with Node and Webpack), so I replaced the single SASS file with a comparable CSS module.

Also, while I've included a storybook-msw addon, it's not really being used right now because there isn't much mocking to do yet. But there should be more of this coming up soon.


acharliekelly:
Wow! That is a lot of failing frontend tests!

frontend/.storybook/USAGE.md
```diff
@@ -0,0 +1,91 @@
+# Storybook Usage
```
alexjball, line 1: Can you call out storybook and link to this in the main Readme, near where we talk about tests?

frontend/.storybook/preview.js
```diff
@@ -0,0 +1,39 @@
...
+
+export const parameters = {
+  actions: { argTypesRegex: "^on[A-Z].*" },
+  controls: {
+    matchers: {
+      color: /(background|color)$/i,
+      date: /Date$/
+    }
+  },
+  // decorators: [
```
alexjball, line 24: Our jest tests and the browser app use helpers/providers.tsx to provide auth context. To simplify story development maybe we can use that here as well as using a formprovider to give the form ctx?

frontend/.storybook/preview.js
```diff
@@ -0,0 +1,39 @@
+import { RouterContext } from "next/dist/shared/lib/router-context"
+// import { FormProvider, useForm } from "react-hook-form"
+// import { AuthProvider } from "../helpers"
+// import { addDecorator } from "@storybook/react"
+// import { initializeWorker, mswDecorator } from "msw-storybook-addon"
+
+// initializeWorker()
+// addDecorator(mswDecorator)
+
+// don't run in production or during build
```
alexjball, line 10: Not sure how storybook msw addon works but we should be able to use the useMockServiceWorker hook to set it up, as we do in app.tsx. you would call the hook in the render fn of the global decorator. It uses the mock if the env is development, not sure what storybook sets

https://github.com/codeforboston/police-data-trust/blob/main/frontend/helpers/api/mocks/browser.setup.ts

frontend/helpers/context.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { ComponentProps, FC } from "react"
+import { AuthProvider } from "."
+import { FormProvider, useForm, useFormContext } from "react-hook-form"
+
+export const combineComponents = (...components: FC[]): FC => {
```
alexjball, line 5: Is this in use?

frontend/package.json
```diff
@@ -12,7 +12,9 @@
     "test:update-snapshots": "jest --updateSnapshot",
     "watch": "jest --watch",
     "fix": "next lint --fix; prettier --write .",
-    "check-types": "tsc --noEmit"
+    "check-types": "tsc --noEmit",
```
alexjball, line 15: How did you set this up initially? Storybook init script? I'm surprised some of the transforms (sass) don't work

alexjball (changes requested):
Above and beyond! Thanks so much for adding stories as well as setup! 

Looks like checks are failing because the package lock needs updating (npm i).

Does npm run check-types catch typescript errors in stories as well?

frontend/helpers/context.tsx
acharliekelly: Oh right. I was going to use it to combine Providers, only I couldn't get it to work. I couldn't get any global context decorators to work - I tried wrapping the FormProvider in the AuthProvider, or vice versa. They both worked within each story decorator, but I couldn't get anywhere with them in .storybook/preview.js.

frontend/.storybook/preview.js
acharliekelly: the addon is similar and looks fun, only I didn't wire it up because there isn't really anywhere to put it - nothing takes mock data yet. but as soon as it does!

frontend/package.json
acharliekelly: yeah, me too - there's even a preset for Sass. but it kept blowing up - not just failing to display that one story (I think it was PrimaryButton) but breaking the Storybook build. apparently there were incompatibilities between versions of node, sass, and... something else, i can't remember. I bet we can get it working eventually, but for the short term, it seemed easier to just rewrite that one file as css and jettison the sass addon

acharliekelly: also, yes, I used `npx sb init` to start it up. but it was a little confusing because it was my third attempt, and I was pulling files in from two earlier branches, so it's possible it got a little jumbled

frontend/.storybook/preview.js
acharliekelly: sounds great! and i'm sure that's how it's _supposed_ to work, because i've read _extensive_ documentation over the past week. have not had much success with global decorators so far, but maybe it was the package-lock? i'll give it another go

acharliekelly: AHA! Ok, have not solved anything, per se, but did finally notice the crucial difference between components that render successfully in storybook and those that blow up: `requireAuth`! Wrapping the story in the `AuthProvider` decorator will give it access to `useAuth`, but if the component is inside of `requireAuth` then something else needs to happen besides just providing context. No idea what, but I'm sure it's something really simple, once you know what you're doing...

frontend/.storybook/USAGE.md
acharliekelly: Yes!

frontend/helpers/context.tsx
acharliekelly: Update: I removed the context.tsx file, and instead changed providers.tsx so it wraps children in FormProvider as well as AuthProvider

alexjball: I think we need to put the global decorators in a separate export in preview.js rather than as part of parameters. Also, we should put the formProvider in preview rather than providers since the formprovider is not global in the real application, which also uses providers. I am able to provide auth and form contexts to all stories with this snippet (we would then remove the decorators from individual stories):

```javascript
// preview.js
export const decorators = [
  (Story) => (
    <Providers>
      <FormProvider {...useForm()}>
        <Story />
      </FormProvider>
    </Providers>
  )
]
```

acharliekelly: That looks great, and makes a lot more sense than having to write all that code into each story! (I just couldn't ever get global decorators to work.)

alexjball (approved):

Issue #168, opened 2021-09-30
[FEATURE] secondary button class


acharliekelly:
Need a CSS class added to `styles/custom.css` that looks like `.primaryButton`, but with reversed colors:

![secondary-button](https://user-images.githubusercontent.com/46508773/135374004-bd1ce16c-3916-4368-a6d3-335db281e212.png)

acharliekelly:
closed by #171

PR #171, opened 2021-10-01
Feat secondary button

acharliekelly:
Issue #168: Secondary Button class, available as CSS globally, that looks like the Primary button only inverted (dark blue text on white)


alexjball (approved):

acharliekelly:
our long-awaited white buttons are finally here!

PR #172, opened 2021-10-01
Fixed Edit Profile

### LINKED ISSUE #166: [BUG] Edit Profile Info causes unhandled error
On the Profile Info page (`/profile`), there's an Edit button to allow you to update your info. 

profile-edit

Instead, it causes an unhandled error. 

profile-edit-error

Probably related to react-hook-form interaction. Also, there may not have been a snapshot of the edit info page.

acharliekelly:
Closes #166. Turns out it was just missing a FormProvider, but went ahead and restructured all the Profile components so they resemble the rest of the project, with components only grouped together if they need to be. Also, the Profile is now getting data (at least user data) from the auth context instead of loading it independently from some JSON file. So it's still mock data, but at least it's the same mock data as everything else, and will therefore be able to switch to real data as soon as that becomes available.


frontend/compositions/profile-type/index.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from "react"
+import Link from "next/link"
+import { profileTypeContent, publicUser, UserDataType } from "../../models/profile"
+import { useAuth, requireAuth } from "../../helpers"
+import styles from "./profile-type.module.css"
+
+export default requireAuth(function ProfileType() {
```
alexjball, line 7: The requireAuth call should wrap the whole Profile page component, so please move to `pages/profile/index.tsx`.

alexjball (approved):
LGTM except for the requireAuth

JMStudiosJoe:
just need to fix conflicts then good to merge :)

frontend/compositions/profile-type/index.tsx
acharliekelly: got it, thanks

acharliekelly:
All set! Even the stories all work now!

alexjball (approved):

acharliekelly:
YES!!

alexjball (approved):

PR #174, opened 2021-10-05
Adds bubble chart

sashamaryl:
- Add bubble chart
- Adds reusable chart key
- Creates visualizations folder. Reorganized visualization routines to share code between bubble chart and map


acharliekelly (commented):
Awesome! I don't understand much of it, but it looks great! (Except for the CI checks, which I also don't understand, but which somebody is probably working on.)

Issue #156, opened 2021-09-20
[FEATURE] Create a /logout page


alexjball:
- Create a page at the route `/logout` that logs the user out and redirects them to the login page. The logout tab directs to that URL
- The logout page can be a blank page or show a loading element like a spinner
- Remove the now-obsolete logout button next to donate

Refer to the `helpers/auth` module.

PR #175, opened 2021-10-06
Logout workflow

ascampos:
#156 

logout

### DIFF
--- a/frontend/pages/logout/index.tsx
+++ b/frontend/pages/logout/index.tsx
@@ -7,7 +7,7 @@ import { LogoSizes } from "../../models"
 import { useAuth } from "../../helpers"
 
 export default function Logout() {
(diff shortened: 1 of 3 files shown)


frontend/pages/logout/index.tsx
```diff
@@ -0,0 +1,29 @@
+import React from "react"
+import { Layout } from "../../shared-components"
+import { Logo as NPDCLogo } from "../../shared-components"
+import styles from "../../compositions/dashboard-header/dashboard-header.module.css"
+
+import { LogoSizes } from "../../models"
+import { useAuth } from "../../helpers"
+
+export default function Logout() {
```
alexjball, line 9: Please wrap this in requireAuth, which will redirect to /login after logout is successful. Check out pages/search/index.tsx

alexjball (changes requested):

PR #193, opened 2021-10-15
Results data table

### LINKED ISSUE #188: [FEATURE] Update search results data table
[FIGMA design](https://www.figma.com/file/TM86P6ePUar5g24pk4h6Du/NPDC-2?node-id=1623%3A2957)
(modified from #179)

results-data-table

The Search Results data table needs to be updated so it will work with API and/or mock incident data.

**Requirements**
- Tooltips on column headers (see design)
- Filterable by column
- Format date field (input from API will be UNIX timestamp)
- Format officer names (data type is `string[]`, needs to cut off if list is too long to fit in table)
- Button to display Full Incident Record for each row

acharliekelly:
This PR has a couple of issues that require input and/or assistance. It's supposed to fix #188, updating the Results table on the Search page. It also updates the Saved Results and Saved Searches tables, even though no one knows what the second one is for and it will probably be dropped from MVP. All the tables have been using react-table, but with fake data written in, so there was no way to use API data. With this PR, the tables now... continue to use fake data, but it might be easier to change now?

The main thing I haven't figured out is the table header. I don't have filtering working yet, and there are a bunch of style problems. I made tooltips for the columns with extra info, but I can't figure out the transparency - you can see the map right through them. Also, there's problems with column width, especially if you sort the columns - the sort arrows make the width jump around madly.

But now you can deploy it and see my mock data! (It's generated fake incident data, merged with a dataset of the 2019 Grammy winners, because random fake data gets boring quickly.) 

(Also, I should really put the mock data wherever mocks are supposed to go, so they don't clutter the build.)

### DIFF
--- a/frontend/compositions/search-results/index.tsx
+++ b/frontend/compositions/search-results/index.tsx
@@ -160,7 +160,7 @@ export default function SearchResultsTable({ incidents = mockData }: SearchResul
               {row.cells.map((cell) => {
                 if (cell.column.id === "id") {
                   return (
-                    <td>
+                    <td key={cell.getCellProps().key}>
                       <FontAwesomeIcon
                         className={actionBtn}
                         title="Save Record"
--- a/frontend/tests/snapshots/__snapshots__/search.test.tsx.snap
+++ b/frontend/tests/snapshots/__snapshots__/search.test.tsx.snap
@@ -663,7 +663,7 @@ exports[`renders Dashboard correctly 1`] = `
           </td>
           <td>
             <svg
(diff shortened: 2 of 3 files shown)


acharliekelly:
The snapshots appear to be failing because of the generated `id` attribute of a FontAwesome icon, which is brand new every time it renders. So I can run `update-snapshots`, and then run `test` immediately after, and it still fails. Does anyone have thoughts on this?

alexjball:
Hey @acharliekelly where can I view your changes in the firebase preview? Or do I need to use storybook?

Issue #131, opened 2021-09-02
[FEATURE] create Search Panel component


acharliekelly:
Search panel component, to appear on the side of the Dashboard page:

![search-panel](https://user-images.githubusercontent.com/46508773/131792082-7c45d8ba-35a8-49ef-90bb-72928e727d21.png)

**Requirements**
- Button to switch between Incident and Officer search
- Location (text) - both types of search
- Incident Type field (drop-down) - Incident only
- Date/Time (pop-up calendar input) - Incident only
- Officer name (text) - Officer only
- Badge number (text) - Officer only

ascampos:
As discussed, we will change "Incident Type" to "Keywords"

PR #206, opened 2021-11-02
Search

ascampos:
Screen Shot 2021-11-02 at 6 55 40 PM

Screen Shot 2021-11-02 at 6 57 06 PM


ascampos:
https://github.com/codeforboston/police-data-trust/issues/131

ascampos:
@alexjball I addressed your comments, but still not sure why there's such a large diff on yarn.lock im using `v1.22.10`

alexjball (approved):
🔥 ! Most folks use npm which tries to update yarn.lock, but apparently not well enough!

Issue #208, opened 2021-11-03
[FEATURE] Expand Tooltip options


acharliekelly:
Need a tooltip, or something very similar to a tooltip, for [column headers](https://www.figma.com/file/TM86P6ePUar5g24pk4h6Du/NPDC--2?node-id=1623%3A3505) that need further explanation, such as Date/Time and Use of Force. The trigger element should be different (info icon rather than question), and the expanded element should allow JSX rather than just a string, since the Use of Force header contains bold text and an `ExternalLink`.

PR #219, opened 2021-12-07
Updates popups and tooltips

sashamaryl:
adds popups for viewing detail about map features
updates tool tip to accept JSX content in addition to strings 
sets up tool tip in Primary input so can be used in Search panel &+
closes issue #208


frontend/compositions/visualizations/map/BaseMap.tsx
```diff
@@ -16,101 +12,70 @@ import { FakeData } from "../utilities/chartTypes"
...
-
-  const svg = d3.select(baseMapRef.current)
+  const valueScale = scaleLinear().domain([0, 5]).range([0, 1])
+  const colorGradient = interpolate("#7caed7", "#303463")
 
+  const svg = select(baseMapRef.current)
   useEffect(() => {
-    if (!geoData) return
-
+    // if (!geoData) return
```
JMStudiosJoe, line 29: can we remove the commented out code?

frontend/compositions/visualizations/map/Map.tsx
```diff
@@ -43,88 +61,73 @@ export default function Map() {
...
-  }, [])
-
-  const handleMouseEnter = useCallback(() => {
-    addToolTip()
-  }, [addToolTip])
+  const handleMapDoubleClick = useCallback(
+    (event: MouseEvent) => {
+      event.stopPropagation()
+      event.stopImmediatePropagation()
+      console.log("doubleclick")
```
JMStudiosJoe, line 101: lets remove the log

frontend/compositions/visualizations/map/marker-layer.tsx
```diff
@@ -1,41 +1,140 @@
...
 export function MarkerLayer(props: MarkerLayerProps) {
   const { markersData } = props
+
+  const handleMouseOver = (event: MouseEvent, markerDescription: MarkerDescription) => {
+    console.log("moue enter marker")
+
+    const x = markerDescription.geoCenter.x + (event.target as HTMLElement).offsetLeft
+    const y = markerDescription.geoCenter.y + (event.target as HTMLElement).offsetTop
+
+    // popup.moveTo({x: x, y: y})
```
JMStudiosJoe, line 26: can we remove the comments & the log?

frontend/compositions/visualizations/map/useSearchData.tsx
```diff
@@ -0,0 +1,31 @@
+import { geoAlbersUsa } from "d3"
+import { Feature, FeatureCollection } from "geojson"
+import { useSearch } from "../../../helpers"
+import { Incident } from "../../../helpers/api"
+
+export default function useSearchData() {
+  const { incidentResults } = useSearch()
+  const projection = geoAlbersUsa()
+    .scale(1300)
+    .translate([487.5 + 112, 305 + 50])
```
JMStudiosJoe, line 10: could we possibly use variable names instead of magic numbers?

frontend/compositions/search/search-panel.tsx
```diff
@@ -58,10 +59,24 @@ export const SearchPanel = () => {
           {incidentSearch ? (
             <div>
               <div>
-                <PrimaryInput inputName={LOCATION} />
+                <PrimaryInput
+                  inputName={LOCATION}
+                  tooltipProps={{
+                    type: TooltipTypes.USEFORCE,
+                    icon: TooltipIcons.INFO,
+                    iconSize: "xs"
```
srsexton94, line 67: Wondering if these have to be props? It seems only the type changes?

Could we map to these based on the input name within the PrimaryInput component?

frontend/compositions/visualizations/map/BaseMap.tsx
```diff
@@ -16,101 +12,70 @@ import { FakeData } from "../utilities/chartTypes"
...
+    //   .attr("in", "colorMatrix2")
+    //   .attr("in2", "blurFilter2")
+    //   .attr("operator", "out")
+    //   .attr("result", "compositedStroke")
+
+    // strokeShape
+    //   .append("feBlend")
+    //   .attr("in", "colorMatrix2")
+    //   .attr("in2", "compositedStroke")
+    //   .attr("mode", "multiply")
```
srsexton94, line 75: If we don't need this anymore we should remove it! Big comment critic here sorry :)

frontend/compositions/visualizations/map/BaseMap.tsx
```diff
@@ -120,19 +85,30 @@ export default function BaseMap(props: BaseMapProps) {
...
+  return (
+    <svg id="map" viewBox={`0, 0, 1200, 700`} ref={baseMapRef}>
+      <rect
+        height={100}
+        width={100}
+        x={100}
+        y={100}
+        fill={"yellow"}
+        onClick={() => {
+          console.log(reload)
```
srsexton94, line 108: remove console.log?

frontend/compositions/visualizations/map/BaseMap.tsx
```diff
@@ -120,19 +85,30 @@ export default function BaseMap(props: BaseMapProps) {
...
-  }, [data, valueScale, path, colorGradient, geoData, svg])
-
-  return <svg id="map" viewBox={`0, 0, 1200, 700`} ref={baseMapRef}></svg>
+  }, [valueScale, path, colorGradient, geoData, svg])
+
+  return (
+    <svg id="map" viewBox={`0, 0, 1200, 700`} ref={baseMapRef}>
+      <rect
+        height={100}
+        width={100}
```
srsexton94, line 103: do we want to do this through CSS instead?

frontend/compositions/visualizations/map/Map.tsx
```diff
@@ -28,12 +37,21 @@ export default function Map() {
   }, [])
 
   const path = useMemo(() => geoPath(projection), [projection])
-
-  const svgElement = select("#map-svg") as D3CallableSelectionType
+  // const svgElementRef = useRef(select("#map-svg") as D3CallableSelectionType)
+  // const svgElement = select("#map-svg") as D3CallableSelectionType
```
srsexton94, line 41: remove commented code?

srsexton94: & on line 53

frontend/compositions/visualizations/map/Map.tsx
```diff
@@ -135,57 +138,58 @@ export default function Map() {
...
+      {/*ts-nocheck*/}
       <div id="map-wrapper" className={styles.mapWrapper}>
         <svg
           id="map-svg"
           role="img"
           aria-labelledby="title desc"
           className={styles.mapData}
           viewBox={`0, 0, 1200, 700`}>
-          <title>US Map Graphic</title>
+          {/* <title>US Map Graphic</title> */}
```
srsexton94, line 185: remove commented out code here

frontend/compositions/visualizations/map/marker-layer.tsx
```diff
@@ -1,41 +1,140 @@
...
+    transformScale
+  } = props
+
+  const coordinate: Coord = markerDescription.geoCenter
+
+  return (
+    <g
+      type={type}
+      {...props}
+      onClick={() => console.log(markerDescription.label)}
```
srsexton94, line 93: remove console.log/method?

frontend/compositions/visualizations/popUps/popUpComp.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <div
+      id={"popup"}
+      className={classNames}
+      style={style}
+      aria-label={ariaLabel}
+      aria-describedby={popUpBodyId}
+      role="tooltip">
+      <header id="head" className={styles.popUpHead}>
+        {headerText}
+      </header>
```
srsexton94, line 33: We won't want to use `header` here - that is a page landmark element that needs to be a direct sibling element to our `main` tag (like the header of the whole website/page)

Instead, this should probably be a "heading" element (h1-6) - which need to follow chronological order, not skipping a number, top-down on the page

frontend/compositions/visualizations/popUps/popUpComp.tsx
```diff
@@ -0,0 +1,39 @@
...
+  location?: Coord
+}
+
+export default function PopUp({ shouldShowPopUp, headerText, bodyText, location }: PopUpProps) {
+  const classNames = `${styles.popUp} ${shouldShowPopUp ? styles.popUpOpen : styles.popUpClosed}`
+
+  const ariaLabel: string = `Detailed information about ${headerText}`
+  const popUpBodyId: string = `${headerText}Details`
+
+  const style = {
```
srsexton94, line 19: Could we do this through a css file instead?

frontend/compositions/visualizations/map/marker-layer.tsx
```diff
@@ -1,41 +1,140 @@
...
+    />
+  )
+}
+
+export function SquareMarker(props: { position: Coord; fill?: string }) {
+  const {
+    position: { x, y },
+    fill
+  } = props
+  const r = 20
```
srsexton94, line 128: Could we use a more informative variable name?

frontend/compositions/visualizations/map/marker-layer.tsx
```diff
@@ -1,41 +1,140 @@
...
+      fill={fill || "var(--darkBlue)"}
+    />
+  )
+}
+
+export function SquareMarker(props: { position: Coord; fill?: string }) {
+  const {
+    position: { x, y },
+    fill
+  } = props
```
srsexton94, line 127: This can be destructured directly in the function parenthesis - and the type can be factored out into an interface for brevity/give us an opportunity to name it informatively!

srsexton94: Same seemingly with the other places we're taking props in this file!

frontend/compositions/visualizations/map/marker-layer.tsx
```diff
@@ -1,41 +1,140 @@
...
+      viewBox="0, 0, 1200, 700"
+      id={"marker-layer"}
+      className={styles.markerLayer}
+      height={"100%"}
+      width={"100%"}>
       {markersData &&
         markersData.map((c, i) => {
-          return <CircleMarker markerDescription={c} key={`${i}${c.dataPoint}`} />
+          return (
+            <>
```
srsexton94, line 51: Do we need the empty container? Since Marker is a single element?

frontend/compositions/visualizations/map/marker-layer.tsx
```diff
@@ -1,41 +1,140 @@
...
+  const handleMouseLeave = () => {
+    console.log("mosue leave makrer")
+    // popup.hide()
+  }
+
   return (
-    <svg viewBox="0, 0, 1200, 700" className={styles.markerLayer} height={"100%"} width={"100%"}>
+    <svg
+      viewBox="0, 0, 1200, 700"
+      id={"marker-layer"}
```
srsexton94, line 44: I don't think we need the brackets here (static value, yeah?)

frontend/compositions/visualizations/map/marker-layer.tsx
```diff
@@ -1,41 +1,140 @@
...
+    // popup.setHeader(markerDescription.label)
+    // popup.setBody("this is a marker popup")
+    // popup.show()
+  }
+
+  const handleMouseMove = (event: MouseEvent) => {
+    event.stopPropagation()
+  }
+
+  const handleMouseLeave = () => {
```
srsexton94, line 36: I'd say remove the console.log/commented out code which leads me to - do we need this method?

frontend/compositions/visualizations/map/Map.tsx
```diff
@@ -198,3 +202,11 @@ export default function Map() {
     <div></div>
   )
 }
+
+export const BodyComp = (bodyContent: string | any) => {
```
srsexton94, line 206: I'm having trouble finding where this is used - seems a bit unnecessary to factor this out given its simplicity, and the div is probably unnecessary - could we potentially just put a paragraph with the body content wherever this is called?

frontend/compositions/visualizations/map/Map.tsx
```diff
@@ -43,88 +61,73 @@ export default function Map() {
...
-    [path]
+    [getHeightAndWidth, path]
   )
 
   const moveMap = useCallback(
     (transform: ZoomTransform, transition = true, pointer?: PointCoord) => {
       const transformStr = transform.toString()
-      if (transformStr.includes("NaN")) return
+      if (transformStr.includes("NaN")) {
+        return
```
srsexton94, line 71: SRP-wise I'm generally a fan of avoiding null/duplicate return statements in any given function - typically you can just check the opposite case before calling the rest of the code (either by wrapping it in the conditional or more ideally checking the condition before invocation) Nonblocker tho!

frontend/compositions/visualizations/map/BaseMap.tsx
```diff
@@ -120,19 +85,30 @@ export default function BaseMap(props: BaseMapProps) {
...
-      })
-      // .attr("pointer-events", "all")
+      .attr("fill", (d: Feature) => "lightblue")
+      .attr("stroke", "white")
       .attr("cursor", "pointer")
+      .attr("pointer-events", "all")
 
     return () => {
-      defs.remove()
+      // defs.remove()
```
srsexton94, line 94: remove commented out code

frontend/compositions/visualizations/map/BaseMap.tsx
```diff
@@ -16,101 +12,70 @@ import { FakeData } from "../utilities/chartTypes"
...
+
+    // strokeShape
+    //   .append("feBlend")
+    //   .attr("in", "colorMatrix2")
+    //   .attr("in2", "compositedStroke")
+    //   .attr("mode", "multiply")
 
     const paths = svg.append("g").attr("id", "paths")
+
     /* SVG Body */
```
srsexton94, line 79: even here I'd be inclined to remove the comment - I think we could capture what this comment is communicating by naming "paths" a little more informatively!

frontend/compositions/visualizations/map/BaseMap.tsx
```diff
@@ -1,10 +1,6 @@
-import * as d3 from "d3"
+import { geoPath, scaleLinear, interpolate, select } from "d3"
 import { Feature, FeatureCollection } from "geojson"
 import React, { useEffect, useRef, useState } from "react"
-import * as topojson from "topojson-client"
-import { presimplify, simplify } from "topojson-simplify"
-import { Topology } from "topojson-specification"
-import { FakeData } from "../utilities/chartTypes"
 
 // colors:
```
srsexton94, line 5: I see it's not your code, but if you don't mind scrapping the comments here too it'd make my heart happy 😅

frontend/compositions/visualizations/map/BaseMap.stories.tsx
```diff
@@ -13,61 +13,10 @@ const Template: ComponentStory<typeof BaseMap> = (args) => <BaseMap {...args} />
...
-    UID: 10,
-    state: "15",
-    value: 20
-  }
-]
+const geoData: FeatureCollection[] = []
 
 Default.args = {
-  data: data,
+  geoData: geoData,
```
srsexton94, line 19: Fun fact, if the key & value are the same you don't have to repeat them in JS! 

Always makes my heart happy when I get to use that 😆

srsexton94 (changes requested):
Hey! Only found one particular accessibility concern - the rest is more about comments/clean code. I haven't pulled the branch yet but I'll try to do so soon to check things are tabbable/screen-reader-able & such!

JMStudiosJoe:
does this still need further changes?

frontend/compositions/visualizations/map/useSearchData.tsx
sashamaryl: These numbers are the projection settings that come in the map boundary data, so to me it's clearer/more error-proof not to make variables.  ```projection``` will be a shareable variable soon and will wrap these.

sashamaryl: These numbers are the projection settings that come in the map boundary data, so to me it's clearer/more error-proof not to make variables.  ```projection``` will be a shareable variable soon and will wrap these.

frontend/compositions/visualizations/map/BaseMap.tsx
sashamaryl: this is a chewing-gum fix for a temp bug having to do with loading the temp data.  should be gone when we hook up the real data sources

frontend/compositions/search/search-panel.tsx
sashamaryl: i didn't actually write this -- I'll have to leave it to the search-panel/table person to move this forward.

frontend/compositions/visualizations/map/BaseMap.tsx
sashamaryl: We will need it again -- this is the styling for the map, but it obscures the actual outlines which makes building/fixing other things which interact with those outlines tough, so it's 'off' while in progress

frontend/compositions/visualizations/popUps/popUpComp.tsx
sashamaryl: not that I know of -- the location is passed in as a prop, and transforms overwrite each other so the translate X needs to be stated here, too.

frontend/compositions/visualizations/map/BaseMap.tsx
sashamaryl: this goes along with the styling

frontend/compositions/visualizations/map/BaseMap.tsx
sashamaryl: this comment is marking the text flow of the document itself, to help read/scan the text, not describing the function of the code... 'paths' is the right name for that variable, and different text wouldn't solve the problem anyway-- I think the solution would actually be ... a longer/more informative comment ... ??  xD

sashamaryl:
@JMStudiosJoe @srsexton94 thanks for your comments, y'all!

frontend/compositions/visualizations/map/BaseMap.tsx
JMStudiosJoe: ok fo rme

frontend/compositions/visualizations/map/BaseMap.tsx
JMStudiosJoe: sweet good to know

JMStudiosJoe (approved):

PR #228, opened 2022-02-01
Updates popups and tooltips - Re-Pull

sashamaryl:
This re-introduces the changes from #219 and addresses the comments in that PR. The revert, #227, does not actually remove the reverted commits from the git history, so the only commit listed in this PR is the one addressing comments, even though it reintroduces all the code from 219.


JMStudiosJoe:
think just need to update snapshot tests then can merge :)

PR #230, opened 2022-02-09
site-wide header, with buttons hidden when not logged in

micahgoodman:


JMStudiosJoe:
few things, need to fix the conflicts in the yarn.lock, can delete it and rerun `npm install` or `yarn install` to generate a new one. Prettier might need to be run and fix linting issues, also snap shots need to be updated with `npm run test -u` I think. Be sure to also pull in the latest changes to avoid conflicts.

JMStudiosJoe:
Still seeing issues with tests, should we connect to see the issue?

Issue #278, opened 2023-05-21
[REFACTOR] Update Incident Model


DMalone87:
**Is your feature request related to a problem? Please describe.**
Our current Incident schema is not compatible with the guideline laid out in the [NPDC Incident Model](https://docs.google.com/document/d/1PCsRIKr0PF14Ul_qnem_UrUHz5RA0-5EmP5NIbDDQGA/edit?usp=sharing). Since this is the model that we plan to share with coalition partners who want to submit data to us, it's imperative that we follow it ourselves.

**Describe the solution you'd like**
Update the current incident schema to accord with the proposed incident model.

**Additional Context**
This change is required before we can tackle [Issue #277](https://github.com/codeforboston/police-data-trust/issues/277), Adding a Data Simulation feature.

DMalone87:
Completed by PR #281

PR #281, opened 2023-06-12
Issue #278 - Refactor Database

DMalone87:
Makes a number of adjustments to the Index database in order to bring it into alignment with the [NPDC Incident Model](https://docs.google.com/document/d/1PCsRIKr0PF14Ul_qnem_UrUHz5RA0-5EmP5NIbDDQGA/edit?usp=sharing). 

* Adds a Suspect model that sits between Officers and Incidents. This takes over for the Accusation table.
* Makes Source a parent of Incident rather than the opposite. Also adds a SourceDetails table to Incidents.
* Adds Many to Many Relationships and association tables.
* Slightly adapts the front end to the new models.

Some things that are not part of this PR that need to be taken care of:
* Current changes will break the scrapers.
* UI needs to be revamped to match the new model properties.
* DB Seeds need to be updated.

DB Seeds update will be a fast follow.


alembic/dev_seeds.py
```diff
@@ -71,34 +71,34 @@ def create_source(source):
...
     incident = Incident(
         id=id,
+        source = mpv,
         location=f"Test location {key}",
         longitude=lon,
         latitude=lat,
         description=f"Test description {key}",
         department=f"Small Police Department {key}",
         time_of_incident=f"{date} 00:00:00",
-        officers=[
```
mnuzzose, line 94: There should probably be seeded officer data as well to be able to test the suspect_matches relationship.

backend/database/models/suspect.py
```diff
@@ -0,0 +1,22 @@
...
+    last_name = db.Column(db.Text)
+    race = db.Column(db.Text)
+    gender = db.Column(db.Text)
+    badge = db.Column(db.Text)
+    unit = db.Column(db.Text)  # type?
+    # Note: rank at time of incident
+    rank = db.Column(db.Enum(Rank))
+    star = db.Column(db.Text)  # type?
+    suspected_matches = db.relationship(
+        "Officer", secondary=suspect_officer, backref="suspect_matches")
```
mnuzzose, line 19: The inconsistency between `suspected_matches` and `suspect_matches` has me wondering if this is 1-to-1 where there is an officer that is definitely the suspect or if it's 1-to-many where there are officers who are suspected of being the suspect?

backend/database/models/suspect.py
```diff
@@ -0,0 +1,22 @@
...
+
+class Suspect(db.Model):
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    incident_id = db.Column(db.Integer, db.ForeignKey("incident.id"))
+    first_name = db.Column(db.Text)
+    last_name = db.Column(db.Text)
+    race = db.Column(db.Text)
+    gender = db.Column(db.Text)
+    badge = db.Column(db.Text)
+    unit = db.Column(db.Text)  # type?
```
mnuzzose, line 14: Is this something where an enum would make sense or is there too much variation between departments?

DMalone87: Yeah, too much variation in the badge and unit values. Most of the time, it should be a number. I'm not certain it always will be though.

backend/database/models/suspect.py
DMalone87: Definitely a many to many. Suspect may be disproven later and that would be recorded in this relationship. Also, you'd need to be able to have a new one.

The naming is clunky though; I agree with you.

DMalone87: Changing this to `Perpetrator` per slack discussion.

backend/database/models/_assoc_tables.py
```diff
@@ -0,0 +1,40 @@
...
+agency_officer = db.Table(
+    'agency_officer',
+    db.Column('agency_id', db.Integer, db.ForeignKey('agency.id'),
+              primary_key=True),
+    db.Column('officer_id', db.Integer, db.ForeignKey('officer.id'),
+              primary_key=True),
+    db.Column('earliest_employment', db.Text),
+    db.Column('latest_employment', db.Text),
+    db.Column('badge_number', db.Text),
+    db.Column('unit', db.Text),
```
mnuzzose, line 29: Just thinking about this - should units be a separate table which relates to an agency? This would help ensure that the unit names would be consistent across relationships which would make identifying problematic units more consistent.

backend/database/models/officer.py
```diff
@@ -18,18 +16,17 @@ class Rank(str, enum.Enum):
     CHIEF = "CHIEF"
 
 
-class Officer(db.Model, SourceMixin):
+class Officer(db.Model):
     id = db.Column(db.Integer, primary_key=True)  # officer id
-    incident_id = db.Column(db.Integer, db.ForeignKey("incident.id"))
     first_name = db.Column(db.Text)
```
mnuzzose, line 21: It feels like `first_name`, `last_name`, `race`, `ethnicity`, `gender`, and `date_of_birth` could be pulled into a table called `individual` which `officer` and `victim` can both reference. This would allow for mapping of situations where an officer later became a victim.

`perpetrator` probably shouldn't have a reference to the `individual` table since it's providing information that would be used to identify suspected officers.

DMalone87: Not sure we need that. The goal of the victim table is just to get demographic data, not to identify the victim as an individual. I want to avoid having any sort of database of victims identities.

DMalone87:
Gonna wave this through with the front end fails. The visualization snapshot is a little flakey and it looks like prettier needs an update.

Issue #274, opened 2023-05-10
[BUG] Yellow square is shown behind map


mikeyavorsky:
**Describe the bug**
A yellow square appears behind the map, for logged-in users.

**To Reproduce**
1. Go to '/login'
2. Enter valid account credentials
3. See yellow square up by Oregon state

**Expected behavior**
There should be no yellow square.

**Screenshots**
<img width="2672" alt="Screen Shot 2023-05-09 at 9 14 45 PM" src="https://github.com/codeforboston/police-data-trust/assets/1855512/9f6a3011-cc27-4915-a0f2-6a454b48bcc7">

PR #282, opened 2023-07-03
Bugfix #274 - Fix Tests

DMalone87:
Updates the snapshot to accommodate the front end changes brought in by @rosyntow 's PR.
Also includes prettier re-write for the BaseMap.


DMalone87:
Merging with the failed test; that build process is deprecated.

PR #292, opened 2023-08-02
284 add contributor role

mnuzzose:


backend/database/models/organization.py
```diff
@@ -0,0 +1,14 @@
...
+from backend.database.models._assoc_tables import organization_user
+
+
+class Organization(db.Model, CrudMixin):
+    id = db.Column(db.String, primary_key=True)
+    name = db.Column(db.Text)
+    url = db.Column(db.Text)
+    contact_email = db.Column(db.Text)
+    reported_incidents = db.relationship(
+        'Incident', backref='organization', lazy="select")
```
DMalone87, line 11: The backref here should still be source.

backend/database/models/incident.py
DMalone87, line 1: The relationship of an organization to an incident is still going to be that of a source -> Incident. The backref on org and the labelling here should stay mostly the same.

backend/database/models/incident.py
```diff
@@ -59,10 +59,10 @@ class Incident(db.Model, CrudMixin):
     """The incident table is the fact table."""
 
     id = db.Column(db.Integer, primary_key=True, autoincrement=True)
-    source_id = db.Column(
```
DMalone87, line 62: These should be reverted.

backend/database/models/incident.py
```diff
@@ -131,8 +131,8 @@ def __repr__(self):
     # case_id = db.Column(db.Integer)  # TODO: foreign key of some sort?
 
 
-class SourceDetails(db.Model):
-    id = db.Column(db.Integer, primary_key=True)  # source details id
+class OrganizationDetails(db.Model):
```
DMalone87, line 134: This is still going to be SourceDetails.

DMalone87 (changes requested):

backend/database/models/organization.py
mnuzzose: If we're renaming `Source` to `Organization` on the back end for consistency with the front end, the database should also reflect that for consistency.

DMalone87:
Gonna go ahead and merge this and make the changes after.

PR #310, opened 2023-12-18
Create Partners Endpoint

hrauniya:
Added some code to existing create partners endpoint to handle existing instances, status codes/response, and validation

New organization/partner cannot be created if an organization already exists in db with same url or email. Partners with same name allowed, but since url/emails are unique I made it so that new partner to be added cannot have the same url/email as existing ones.


DMalone87:
Closing this since it was merged in through another PR

Issue #313, opened 2024-01-03
[FEATURE] Allow Statewide Officer IDs to be recorded in the Index.


DMalone87:
**Is your feature request related to a problem? Please describe.**
One critical function of the system is identifying officers even as they move between different agencies during their careers. While our current schema allows the tracking of badge numbers, these are not enough on their own. Badge numbers are unique within a single department at a particular time, but may be shared by multiple officers over time or when considering the nation as a whole.

To deal with this, some states employ a statewide ID system that allows for tracking individual officers within the state. For example, New York uses Tax ID numbers and the Texas Commission of Law Enforcement (TCOLE) assigns each officer a PID number. We should enable our database to properly track these numbers.

**Describe the solution you'd like**
We should add a field to the officer and perpetrator schemas that allow us to track Statewide officer IDs. We should consider that the same officer may have multiple state IDs if they have been employed in multiple states. 

**Additional context**
The first set of counties we'll be releasing the Index for will be in Texas and New York. This needs to be completed before the web scraper micro service is implemented.

PR #314, opened 2024-01-04
Feature #313 - Add Statewide IDs

DMalone87:
Adds a StateID model as a child of the officer table. This allows the index to track statewide Officer IDs such as Tax ID or TCOLE PID numbers.

### DIFF
--- a/backend/database/models/officer.py
+++ b/backend/database/models/officer.py
@@ -16,6 +16,23 @@ class Rank(str, enum.Enum):
     CHIEF = "CHIEF"
 
 
+class StateID(db.Model):
+    """
+    Represents a Statewide ID that follows an offcier even as they move between
+    law enforcement agencies. for an officer. For example, in New York, this
+    would be the Tax ID Number.
+    """
+    id = db.Column(db.Integer, primary_key=True)
+    officer_id = db.Column(
+        db.Integer, db.ForeignKey("officer.id"))
+    id_name = db.Column(db.Text)  # e.g. "Tax ID Number"
+    state = db.Column(db.Text)  # e.g. "NY"
+    value = db.Column(db.Text)  # e.g. "958938"
+
+    def __repr__(self):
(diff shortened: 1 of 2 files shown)


backend/database/models/officer.py
```diff
@@ -16,16 +16,30 @@ class Rank(str, enum.Enum):
...
+
+
 class Officer(db.Model):
     id = db.Column(db.Integer, primary_key=True)  # officer id
     first_name = db.Column(db.Text)
     last_name = db.Column(db.Text)
     race = db.Column(db.Text)
     ethnicity = db.Column(db.Text)
     gender = db.Column(db.Text)
-    # Note: rank at time of incident
```
DMalone87, line 26: Rank and Star are removed since those are dependent on the Law Enforcement agency the officer is employed by and are not unique to the officer in all cases. Those values can still be found in the association table for Officers and Agencies.

backend/database/models/officer.py
```diff
@@ -16,16 +16,30 @@ class Rank(str, enum.Enum):
...
+    """
+    Represents a Statewide ID that follows an offcier even as they move between
+    law enforcement agencies. for an officer. For example, in New York, this
+    would be the Tax ID Number.
+    """
+    id = db.Column(db.Integer, primary_key=True)
+    officer_id = db.Column(
+        db.Integer, db.ForeignKey("officer.id"))
+    id_name = db.Column(db.Text)  # e.g. "Tax ID Number"
+    state = db.Column(db.Text)  # e.g. "NY"
```
RyEggGit, line 29: Does this make more sense as an enum so we don't have conflicting states like NY and New York?

DMalone87: Yes, great point

DMalone87: Updated.

RyEggGit (approved):

PR #316, opened 2024-01-11
Scraper

RyEggGit:
# Scraper Update 🕷️

Sorry for the big PR; got carried away having to convert the code from one repo to another. For simplicity, this scraper code runs asynchronously with a cronjob that calls the flask command `sracpe-v2`. This iteratively crawls through the 50-a website, gathers as much information as it can (still working on agencies), and downloads the CSV's from NYPD. I then check if the information is in our scraping cache; if not, we add it to the database. The incident schemas are a bit whack and may create duplicated information based on whether the officer is a perpetrator, participant, etc., but they should get the job done for now.


docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
   api:
     build:
       context: .
-      dockerfile: ./backend/Dockerfile
+      dockerfile: ./backend/Dockerfile.cloud
```
mikeyavorsky, line 28: This change should be reverted, breaks `docker-compose build` on ARM Macs.

mikeyavorsky (changes requested):

DMalone87:
Something to consider... we'll need to add some logic to handle 429 responses. Right now, it just continues making requests, but ideally, it should suspend the process and pick it up after a time. Even better it could track the occurrences of these requests and collect data on them for review.

backend/scraper/__init__.py
aliavni, line 1: Hi folks, I am a new contributor to this repo. This work looks nice. Here is my five cents:

1. We may want to have scraping-related work outside this repo. We already have quite a bit going on here. Having a separate repo would be easier in the long for example for deployments and requirements management.
2. I think we can eliminate the redis portion. Once scrapers generate data files, we can load the new data using upserts.
3. There are Python packages dedicated to scraping. One of the great ones is [Scrapy](https://docs.scrapy.org/en/latest/index.html). It is a pretty mature package with all kinds of conveniences such as handling retries, user agents etc.

aliavni: Here is an example scraper with Scrapy I built before https://github.com/aliavni/drugbank-scraper

PR #318, opened 2024-01-18
Ryan/epic2/userstory6

RyEggGit:
Ticket: https://www.notion.so/developforgood/b3a0d3450c3f4440bed354fcbbbbd455?v=99a42947c6f9479ca891d4f0ee3462e1&p=a8b09e0582104c08a4ad4026363bb1bd&pm=s

### DIFF
--- a/alembic/dev_seeds.py
+++ b/alembic/dev_seeds.py
@@ -6,6 +6,7 @@ from backend.database.models.perpetrator import Perpetrator
 from backend.database.models.partner import Partner
 from backend.database.models.use_of_force import UseOfForce
 
+
 def create_user(user):
     user_exists = (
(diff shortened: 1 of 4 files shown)


DMalone87:
So, the test passes fine, but I'm getting an error when I try to run Create on Postman. I think this is a preexisting error though so not your fault. It seems to be trying to reuse the Partner ID "1" which is probably why it doesn't fail the test and why I missed it earlier.

You may run into issues when trying to build on the API. But having something requesting against it will make the troubleshooting faster so I'm inclined to push this through. I'll leave it up to you all if you want to press ahead.

DMalone87 (dismissed):
Approved with some reservation.

DMalone87 (approved):

PR #319, opened 2024-01-20
Epic 2 APIs

hrauniya:
1)Inviting users to NPDC
Ticket:
https://www.notion.so/developforgood/b3a0d3450c3f4440bed354fcbbbbd455?v=99a42947c6f9479ca891d4f0ee3462e1&p=1dedd2178ba94239963285adddbc672b&pm=s

The configuration settings for the mail server needs to be changed in the dotenv depending on if we're using Amazon Simple Email Service or some other tool. I was using mailtrap.io for testing.

2)Users can join/leave organization
Ticket:
https://www.notion.so/developforgood/b3a0d3450c3f4440bed354fcbbbbd455?v=99a42947c6f9479ca891d4f0ee3462e1&p=9f0b34313c434b90b97c296d57cbe5b3&pm=s

3)Remove users from org/withdraw their invitation
Ticket:
https://www.notion.so/developforgood/b3a0d3450c3f4440bed354fcbbbbd455?v=99a42947c6f9479ca891d4f0ee3462e1&p=db318391f99849c488c38c0c6610ef2e&pm=s

4)Change roles of users in a partner org
Ticket:
https://www.notion.so/developforgood/b3a0d3450c3f4440bed354fcbbbbd455?v=99a42947c6f9479ca891d4f0ee3462e1&p=3cbd315ee9144897b36d476f7832dae8&pm=s


backend/routes/partners.py
```diff
@@ -193,3 +203,278 @@ def add_member_to_partner(partner_id: int):
...
+@jwt_required()
+@min_role_required(MemberRole.ADMIN)
+def role_change():
+    body = request.get_json()
+    try:
+        user_found = PartnerMember.query.filter_by(
+            user_id=body["user_id"],
+            partner_id=body["partner_id"]
+            ).first()
+        if user_found:
```
DMalone87, line 425: Let's make sure the user in question isn't also an admin. Admins shouldn't be able to demote another admin.

backend/routes/partners.py
```diff
@@ -193,3 +203,278 @@ def add_member_to_partner(partner_id: int):
...
+        db.session.rollback()
+        return {
+            "status": "Error",
+            "message": "Something went wrong!"
+        }
+    finally:
+        db.session.close()
+
+
+# inviting anyone to NPDC
```
DMalone87, line 284: We should probably just fold this into the existing "Add member" endpoint. Otherwise, we kind of have an end-run around the invitation process.

DMalone87 (changes requested):
Great work! Needs a few changes though.

backend/routes/partners.py
hrauniya: Do you mean shifting the code for the current "invite_user" function/endpoint to  "add_member_to_partner" function/endpoint and deleting the "invite_user" one? Since both of those have the same purpose.

DMalone87: Yes, exactly

hrauniya:
Added changes to admins not being able to demote other admins and moving code from invite_users to "add_members_to_partners".

backend/routes/partners.py
```diff
@@ -140,105 +149,292 @@ class Config:
...
+        db.session.close()
 
-    created = partner_member.create()
+# admin can remove any member from a partner organization
 
-    track_to_mp(
-        request,
-        "add_partner_member",
+
+@bp.route("/remove_member", methods=['DELETE'])
```
DMalone87, line 315: Just like the role change, we shouldn't let an admin forcibly remove another admin without notification.

backend/tests/test_partners.py
```diff
@@ -221,18 +220,22 @@ def test_partner_pagination(client, example_partners, access_token):
     assert res.status_code == 404
```
DMalone87, line 222: We should add tests for the new API endpoints here. We're missing: 

- /invite
- /join
- /leave
- /remove_member
- /withdraw_invitation
- /role_change
- /invitations
- /stagedinvitations

backend/tests/test_partners.py
```diff
@@ -221,18 +220,22 @@ def test_partner_pagination(client, example_partners, access_token):
...
+    #     .first()
+    # )
 
-    assert partner_member_obj.partner_id == created["partner_id"]
-    assert partner_member_obj.user_id == created["user_id"]
-    assert partner_member_obj.role == created["role"]
+    # assert partner_member_obj.partner_id == created["partner_id"]
+    # assert partner_member_obj.email == created["email"]
+    # assert partner_member_obj.role == created["role"]
+    """
```
DMalone87, line 235: No need to worry about testing the mailing functionality here. We can handle that in a seperate test. For the purposes of these tests, just assume the mail feature is working.

DMalone87 (changes requested):
Still a few changes needed on this one. Sorry I didn't catch the missing tests issue earlier.

backend/tests/test_partners.py
hrauniya: I've added tests for every endpoint except /invite. That one requires the mail server to be set up. Also, did not add tests for /invitations, and /staged_invitations as those are for the dev env, and I just wanted endpoints to view the current state of the db. We could remove them if you want.

DMalone87 (dismissed):
Looks good!

hrauniya:
Looks like I have to re-request review for this after fixing some merge conflicts between main and invite_users.

DMalone87 (approved):

DMalone87:
No problem! You're good to go.

PR #323, opened 2024-02-11
create incidents

hrauniya:
create incidents endpoint


DMalone87:
For this change to pass the tests, we'll need to update the test to ensure that the `contributor_access_token` user has the proper permission within the partner orgs that are used for the create incident test. ([See here.](https://github.com/codeforboston/police-data-trust/blob/9bb0ae8e12e52e7f7484f9e552b6c8e2db17461b/backend/tests/test_incidents.py#L54))

Right now, those two elements are mismatched.

PR #326, opened 2024-02-15
officer end point

zeelapatel:
Created an officer search end point by replicating incident search end point 
out put can be changed as which data we want to display in json format


backend/routes/incidents.py
DMalone87, line 1: Let's move these endpoints into a separate file, `officers.py`. It helps to keep things organized. We'll need to add tests as well.

backend/schemas.py
DMalone87, line 1: Looks good!

DMalone87 (changes requested):
Looking good! Just need to make a few adjustments.

backend/routes/incidents.py
zeelapatel: okay I will create a separate file for it.
if there is any other changes please let me know. thank you

DMalone87: We can drop all changes from this file as well since it shouldn't be affected by the Officer endpoints.

frontend/compositions/search-panel/search-panel.tsx
```diff
@@ -47,6 +47,7 @@ export const SearchPanel = (props: SearchPanelProps) => {
     } catch (e) {
       console.error("Unexpected search error", e)
       setErrorMessage("Something went wrong. Please try again.")
+
```
DMalone87, line 50: Let's drop this change as well.

DMalone87:
Thanks, Zeel! There are some style issues, but for this one, I'll go ahead and merge and make the corrections.

PR #340, opened 2024-02-23
WIP: update node to fix firebase gh action

zganger:
- firebase publish failing because it expects node 18 or 20, but is set to version 16.


DMalone87:
Don't worry about this one. We don't actually use that action anymore.

PR #352, opened 2024-03-05
Officer Table Overhaul

DMalone87:
- Adding tests for Officer API endpoints
- Creating association objects for: 
  - Accusation (Officer/Perpetrator) 
  - Employment (Officer/Agency)
- Added new officer endpoints
  - Get all officers
  - Get officer by ID
  - Create Officer


backend/database/models/officer.py
```diff
@@ -86,14 +86,34 @@ def __repr__(self):
         return f"<StateID {self.id}>"
 
 
+class Employment(db.Model):
```
zganger, line 89: You have this defined twice; here and in `employment.py`

DMalone87: Nice catch. Thanks!

backend/routes/officers.py
```diff
@@ -51,19 +55,35 @@ def search_officer():
...
-        if body.officerName:
+        if body.name:
             names = body.officerName.split()
-            first_name = names[0] if len(names) > 0 else ''
-            last_name = names[1] if len(names) > 1 else ''
-            query = Officer.query.filter(or_(
-                Officer.first_name.ilike(f"%{first_name}%"),
-                Officer.last_name.ilike(f"%{last_name}%")
-            ))
+            if len(names) == 0:
```
zganger, line 60: This would be an empty string. We would never accept this case, would we? I think this case and next should have their values incremented. On the `len(names) == 1` case, we would experience an out of bounds exception when indexing `names[1]`.

backend/schemas.py
```diff
@@ -202,6 +208,12 @@ class CreatePartnerMemberSchema(BaseModel):
     is_active: Optional[bool] = True
 
 
+class CreateOfficerSchema(_BaseCreateOfficerSchema, _OfficerMixin):
+    known_employers: Optional[List[CreateEmploymentSchema]]
+    accusations: Optional[List[CreateAccusationSchema]]
+    state_ids: Optional[List[CreateStateIDSchema]]
+
+
```
zganger, line 216: Why do we have two identical schemas here except for optionality (CreateOfficerSchema and OfficerSchema)?

DMalone87: They aren't entirely identical. The create version (`schema_create`) just excludes the ids. There's probably a cleaner way of doing it, but I haven't looked into it. Much of this code was inherited from an earlier cohort who have mostly moved on to other projects or left CfB.

PR #354, opened 2024-03-10
minor correction

zeelapatel:
changed officer_orm_to_json return statement

### DIFF
--- a/backend/routes/officers.py
+++ b/backend/routes/officers.py
@@ -62,7 +62,8 @@ def search_officer():
 
         if body.badgeNumber:
             officer_ids = [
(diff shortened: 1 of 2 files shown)


zeelapatel:
officer search using badgeNumber was not working because of pagination error on query.
I made changes and it works fine!

DMalone87:
Looks like we had the same idea. Fixed the first part of this in PR #352.

Mind rebasing to my branch and adding the BadgeNumber change there? I changed the way the Agency/Officer association is handled.

zeelapatel:
I tested it on postman but it gives me error. I didn't change query but there was some pagination error

PR #364, opened 2024-03-20
Frontend: Profile Page Tabs as URL Params

joshua-rdrgz:
This PR contains some changes that moves the active tab state into the URL, that way the user can navigate to different profile page tabs via the URL, and the app can dynamically switch tabs if it needs to.

# Demonstration
Screenshot 2024-03-20 at 1 31 54 PM
Screenshot 2024-03-20 at 1 31 38 PM
Screenshot 2024-03-20 at 1 32 31 PM


joshua-rdrgz:
# Build Command Replacement

With the implementation of this feature, I had to switch the "/profile" page to an SSR page instead of a static page -- this meant that using `next export` for the build commands was no longer feasible for the entire application (see more [here](https://nextjs.org/docs/messages/gssp-export)).  I've changed the build commands from `next export` to `next start`, which will spin up a NodeJS server capable of SSR for the "/profile" page.  I've also removed `next export` from the `build:ui` command, as adding `next start` to this particular command causes the build tests to never finish (the spun up server will never spin back down).

```json
  "scripts": {
    "build:dev": "env-cmd -f .env.development next build && next start",
    "build:stage": "env-cmd -f .env.staging next build && next start",
    "build:prod": "env-cmd -f .env.production next build && next start",
    "build:ui": "env-cmd -f .env.ui next build",
  },
```

mikeyavorsky:
`&& next start` will break Dockerfile.cloud as written. We need to either background that task or remove `&& next start` from the rest of the commands in package.json, and write a separate `start: next start` command.

frontend/package.json
```diff
@@ -4,10 +4,10 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
-    "build:dev": "env-cmd -f .env.development next build && next export",
-    "build:stage": "env-cmd -f .env.staging next build && next export",
-    "build:prod": "env-cmd -f .env.production next build && next export",
-    "build:ui": "env-cmd -f .env.ui next build && next export",
+    "build:dev": "env-cmd -f .env.development next build && next start",
+    "build:stage": "env-cmd -f .env.staging next build && next start",
+    "build:prod": "env-cmd -f .env.production next build && next start",
```
mikeyavorsky, line 9: Please remove these `&& next start` and add a command `"start": "next start"`, or edit Dockerfile.cloud to add a ` &` to the end of the `npm run build:{...` command.

mikeyavorsky (changes requested):

frontend/package.json
joshua-rdrgz: Do we need to do both of these things?  I've adjusted the build commands to remove `next start` from them -- do I need to add `& npm run start` to the Dockerfile.cloud RUN command as well?

```json
"scripts": {
    "build:dev": "env-cmd -f .env.development next build",
    "build:stage": "env-cmd -f .env.staging next build",
    "build:prod": "env-cmd -f .env.production next build",
    "build:ui": "env-cmd -f .env.ui next build",
    "start": "next start",
  },
```

PR #366, opened 2024-03-27
use unittest to define test cases

zganger:


zganger:
I am abandoning this- further investigation reveals that some changes would be required to the purest fixtures to add them to the test classes, which would be ok, but going up the chain they depend eventually on the `db_session` fixtures provided by flask-sqlalchemy, which I can't edit. For the purposes of a project like this, we are looking at replacing this functionality with custom db connections in a base class or continuing using pytest without the unittest suite. I view this as having too limited a benefit to warrant the change required to undertake the migration.

DMalone87:
Appreciate the effort, Zach!

PR #375, opened 2024-04-18
Rename env variable POSTGRES_PORT -> PGPORT

aliavni:
Postgres requires `PGPORT` environment variable set if we need to run it on a different port. 

This PR 
1. Renames `POSTGRES_PORT` env variable to `PGPORT` so that we can run the app with the postgres port we need. Api service fails to connect to postgres without this change
2. Creates `.env.template` file for easier developer onboarding, and updates README


aliavni:
@DMalone87 one more PR here

DMalone87 (approved):

PR #376, opened 2024-04-23
Officer, Agency combined Search

hrauniya:
-Added alembic migration to set up a Postgres materialized view for the search function to search relevant Officers, Agencies.
-Migration handles insertions, deletions etc. to Agency, Officer Tables, and updates the view necessary for search.
-Search function returns most relevant results according to search terms specified in the function.

Questions:
- Should the API return every attribute of Officer, Agency found through the search or just some?
- Indication necessary as to if search result is either an Officer, or Agency?


.env.save
```diff
@@ -0,0 +1,7 @@
+POSTGRES_USER=postgres
```
hrauniya, line 1: remove this file

backend/routes/officer_agency_search.py
```diff
@@ -0,0 +1,108 @@
...
+class TSRank(GenericFunction):
+    package = 'full_text'
+    name = 'ts_rank'
+    inherit_cache = True
+
+
+bp = Blueprint("search_route", __name__, url_prefix="/api/v1/search")
+"""
+TODO:
+Convert search function to endpoint,
```
DMalone87, line 19: Is there another endpoint you're planning or is this TODO satisfied with this PR?

hrauniya: Yes, the TODO is satisfied. Forgot to remove it.

PR #386, opened 2024-05-15
Build ARM & Intel images for back-end

mikeyavorsky:
- builds and pushes linux/amd64 and linux/arm64 images to ECR
- updates pandas from 1.5.3 to 2.2.2 and installs from wheel to reduce build times


mikeyavorsky:
Doesn't work without arm64 Github Actions runner (currently in private beta?)

backend/Dockerfile.cloud
```diff
@@ -3,14 +3,20 @@
...
 ARG PDT_API_PORT=5000
 
 COPY requirements/prod.txt .
-RUN pip install -r prod.txt
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+RUN arch=$(arch) && \
+	file=pandas-2.2.2-cp312-cp312-manylinux_2_17_${arch}.manylinux2014_${arch}.whl && \
+	url="https://pypi.debian.net/pandas/${file}" && \
+	wget ${url} && \
+	sed -i "s/pandas==1.5.3/${file}/" prod.txt
```
zganger, line 18: Probably ought to update the requirements files as well- both for any dependencies that 2.2.2 expects and so that other dockerfiles and local get the same version for testing.

mikeyavorsky: [Thanks, done.](https://github.com/codeforboston/police-data-trust/pull/386/commits/91635c5981fccb1eaec62636eb042cd03c4f7ecf)

zganger (approved):
:100:

Issue #387, opened 2024-05-22
[FEATURE] Convert the Unit property into a DB Model


DMalone87:
**Is your feature request related to a problem? Please describe.**
Each police officer working within a police department will be associated with a particular unit. For example, NYPD Officer [Kit Yung](https://www.50-a.org/officer/QHZ3) is a Sergeant in the [9th Precinct](https://www.50-a.org/command/9pct).

In our current database structure, a Unit is just a string property of the employment model ([See here](https://github.com/codeforboston/police-data-trust/blob/53fca537f431522780af2ff7e8113fa549304556/backend/database/models/employment.py#L26)). In this model, a unit exists between a single officer and a single agency. This doesn't properly represent a world where many officers can belong to the same Unit. 

**Describe the solution you'd like**
We should add a Unit model as a child of an [Agency](https://github.com/codeforboston/police-data-trust/blob/main/backend/database/models/agency.py). We can then associate the Unit object with the Employment object, properly representing the relationship.

PR #398, opened 2024-06-11
 [FEAT] Convert the Unit property into a db Unit model

RyanPark44:
Converted the unit property, in the Agency Model, into its own Unit Model.
[See issue here](https://github.com/codeforboston/police-data-trust/issues/387)

### Completed
- Created new Unit Model
- Established Unit as a child of agency
- Established an association between Unit and Employment object
- Updated Employment Model's merge_employment_records to account for Unit model
- Updated agencies and officers endpoints to reflect change to merge_employment_records
- Updated schema to include Unit model

### To do next
- Create endpoint for Unit
- Create tests for Unit

### DIFF
--- a/backend/database/__init__.py
+++ b/backend/database/__init__.py
@@ -11,6 +11,7 @@ from backend.database.core import execute_query
 #  Alembic only does a couple models, not all of them.
 
 from .models.agency import *
+from .models.unit import *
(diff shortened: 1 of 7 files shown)


DMalone87 (approved):
Looks good! Thanks, Ryan.

PR #404, opened 2024-06-26
Listen to SQS events for s3 object import from scraper

zganger:


DMalone87 (approved):
Looks good from what I can tell. Let's see if we can get it running on the cloud.

PR #411, opened 2024-10-08
update python

zganger:


zganger:
The issues at hand are primarily sqlalchemy/psycopg issues (sqlalchemy v1.x.x uses psycopg2, which has been replaced by psycopg). This will work after we drop sql- will rebase later.

DMalone87 (approved):

PR #452, opened 2025-06-24
Add unit testing with Vitest

mikeyavorsky:
We are working to add playwright end-to-end testing to our Github Actions workflows. In the meantime, we can add unit testing for front-end components with Vitest and run that on pull requests, similar to what we used to do with Jest.

### DIFF
--- a/.github/workflows/frontend-checks.yml
+++ b/.github/workflows/frontend-checks.yml
@@ -28,3 +28,7 @@ jobs:
       - name: Formatting
         if: always()
         run: npm run check-formatting
+
+      - name: Vitest
+        if: always()
+        run: npm run test:unit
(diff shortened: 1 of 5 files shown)


.github/workflows/frontend-checks.yml
DMalone87, line 1: On my first run, I had to run `npm install` to add Vitest. That shouldn't be required after a rebuild I think... maybe there's a step missing to update something as part of this PR?

DMalone87 (approved):

.github/workflows/frontend-checks.yml
DMalone87: Talked to @signalflowsean and confirmed that this is normal.

PR #461, opened 2025-08-01
Adds pagination to search results

DMalone87:
Adds pagination and a function that can be used to apply the pagination schema to any list of items.


DMalone87:
Using other version

PR #482, opened 2025-09-22
Update User model

DMalone87:
- Added Contact nodes for email, phone numbers, and social media
- Added a user factory method to create users with contacts
- User may have more than one email address
- User may have more than one phone number
- User has a primary email address
- User now has social media information (Twitter handle, Facebook profile, etc)


DMalone87:
Will merge without a migration runner for now to allow progress on blocked tasks. The migration scripts are included though.

PR #515, opened 2026-01-07
Moved pydantic models to dto/ and fixed pagination bug in routes

maxaaronson:


backend/dto/agency.py
DMalone87: Let's group the DTO files based on the way they're grouped in `models` rather than routes. Routes can change and touch multiple domains, but the models grouping will be pretty stable.

DMalone87: Actually, never mind. I see you have some route specific logic included here. This is good.

DMalone87 (approved):

PR #525, opened 2026-01-30
[Feature] Issue 440 fuzzy search

JMStudiosJoe:
added to query level for level of effort is 1 line and goes to each search query in the backend. found details for change from https://stackoverflow.com/questions/71162486/neo4j-full-text-index-terms-found


JMStudiosJoe:
wooooo

